PR #25, opened 2022-12-14
Fix surface muons

MoritzNeuberger:
Overview:
- Changed the generator to inherit from RMGVGenerator
- Added macro option to set sky height in plane sky shape mode
- Added new example based on 02-HPGe to test the surface muon generator
- Added cut production options for mu+ and mu- with the same cut value
- Fixed a bug with the HPGe density (unit was missing)


MoritzNeuberger:
Should be removed now.

MoritzNeuberger:
I should have implemented all the points we talked about earlier.

PR #26, opened 2022-12-15
Implement musun input generator

MoritzNeuberger:
Overview:
- Added class to generate muons using a MUSUN file input
- Added class to read a MUSUN file (singleton) 
- Added example to experiment with the new generator based on 02-HPGe
- Added an example MUSUN input file with 10000 muons (MUSUN_10k_events.dat)


include/RMGReaderMUSUN.hh
```diff
@@ -0,0 +1,43 @@
...
+    int particleID;
+    float energy;
+    float x;
+    float y;
+    float z;
+    float theta;
+    float phi;
+};
+
+class RMGReaderMUSUN {
```
gipert, line 18: Can you investigate if a MUSUN file can be interpreted as CSV, so we can use `G4CsvAnalysisReader` to parse it (see `RMGVertexFromFile`)?

gipert:
What is the status here? Only need to better handle the header as we discussed?

include/ProjectInfo.hh
```diff
@@ -0,0 +1,9 @@
+// other useful GEANT4 exported variables:
```
ManuelHu, line 1: This file should be removed, it has been replaced by the (git-excluded) `RMGConfig.hh`

include/RMGGeneratorMUSUNCosmicMuons.hh
```diff
@@ -0,0 +1,66 @@
...
+
+#include "CLHEP/Units/SystemOfUnits.h"
+#include "G4GenericMessenger.hh"
+#include "G4ParticleGun.hh"
+#include "G4VUserPrimaryGeneratorAction.hh"
+
+
+#include "G4ThreeVector.hh"
+
+#include "ProjectInfo.hh"
```
ManuelHu, line 15: `RMGConfig.hh`

ManuelHu (changes requested):
This file should be deleted (also locally,

ManuelHu:
Now there are some parts of recent `main` branch changes duplicated in this PR... (and `ProjectInfo.hh` is still there, which is the cause for the slim build failures...) Maybe a rebase on the latest main would help :-)

ManuelHu (changes requested):

MoritzNeuberger:
So, a problem with the multithreaded implementation. G4CsvAnalysisReader expects a separate file for each thread. Specifically, it expects them in this format: `FILENAME_nt_MUSUN_t0.csv, FILENAME_nt_MUSUN_t1.csv, ...` where tX is the iterable. Since we already control the input file in the copy step, I guess you could get around this by splitting the initial file into `nThreads` number of sub-files that we feed into the threads.

MoritzNeuberger:
So, there is a problem with the multithreaded implementation. The G4CsvAnalysisReader instance is currently initialized in the `BeginRunAction` of the generator class, so each thread will spawn a separate instance, meaning that each thread will go through all entries in the input file. Could we instead initialize it in the user action?

MoritzNeuberger:
At this point, I am considering testing whether a Singelton with some mutex might be the way to go.

MoritzNeuberger:
As this point I am thinking of working with a singelton and mutex similar to [this](https://github.com/Geant4/geant4/blob/master/examples/extended/runAndEvent/RE05/src/RE05PrimaryGeneratorAction.cc) Geant4 example.

MoritzNeuberger:
In the example I now get this seg violation:

```
 *** Break *** segmentation violation
WARNING - Attempt to delete the physical volume store while geometry closed !
WARNING - Attempt to delete the logical volume store while geometry closed !
WARNING - Attempt to delete the solid store while geometry closed !
WARNING - Attempt to delete the region store while geometry closed !
```

I guess I messed up somewhere.

MoritzNeuberger:
Seems to not throw the seg violation anymore. It was probably related to an automatic delete of the `G4CsvAnalysisReader` at the end of the program, making the manual delete unnecessary?

Anyway, I think the pull request is not feature-complete. What do you guys think?

examples/05-MUSUN/run.mac
```diff
@@ -0,0 +1,16 @@
+# /RMG/Manager/Logging/LogLevelScreen debug
```
ManuelHu, line 1:
```suggestion
# /RMG/Manager/Logging/LogLevel debug
```

examples/05-MUSUN/vis-traj.mac
```diff
@@ -0,0 +1,22 @@
+/RMG/Manager/Logging/LogLevelScreen debug
```
ManuelHu, line 1:
```suggestion
/RMG/Manager/Logging/LogLevel debug
```

src/RMGGeneratorMUSUNCosmicMuons.cc
```diff
@@ -0,0 +1,204 @@
+#include "RMGGeneratorMUSUNCosmicMuons.hh"
+
+#include "RMGVGenerator.hh"
+#include "math.h"
```
ManuelHu, line 4:
```suggestion
#include <filesystem>
#include <math>
#include <vector>
```

src/RMGGeneratorMUSUNCosmicMuons.cc
```diff
@@ -0,0 +1,204 @@
...
+#include "G4ParticleMomentum.hh"
+#include "G4ParticleTypes.hh"
+#include "G4ThreeVector.hh"
+#include "Randomize.hh"
+
+#include "RMGHardware.hh"
+#include "RMGLog.hh"
+#include "RMGManager.hh"
+#include "RMGTools.hh"
+
```
ManuelHu, line 17:
```suggestion
#include "RMGVGenerator.hh"

```

src/RMGGeneratorMUSUNCosmicMuons.cc
```diff
@@ -0,0 +1,204 @@
...
+#include "G4AutoLock.hh"
+#include "G4CsvAnalysisReader.hh"
+
+#include <filesystem>
+#include <vector>
+
```
ManuelHu, line 23:
```suggestion
```

ManuelHu (changes requested):
Some macro commands will not work on current main

src/RMGGeneratorMUSUNCosmicMuons.cc
```diff
@@ -0,0 +1,204 @@
+#include "RMGGeneratorMUSUNCosmicMuons.hh"
+
+#include <filesystem>
+#include <math>
```
ManuelHu, line 4:
```suggestion
#include <cmath>
```

ManuelHu (changes requested):

include/RMGGeneratorMUSUNCosmicMuons.hh
```diff
@@ -0,0 +1,71 @@
...
+    RMGGeneratorMUSUNCosmicMuons();
+    ~RMGGeneratorMUSUNCosmicMuons() = default;
+
+    RMGGeneratorMUSUNCosmicMuons(RMGGeneratorMUSUNCosmicMuons const&) = delete;
+    RMGGeneratorMUSUNCosmicMuons& operator=(RMGGeneratorMUSUNCosmicMuons const&) = delete;
+    RMGGeneratorMUSUNCosmicMuons(RMGGeneratorMUSUNCosmicMuons&&) = delete;
+    RMGGeneratorMUSUNCosmicMuons& operator=(RMGGeneratorMUSUNCosmicMuons&&) = delete;
+
+    void GeneratePrimaries(G4Event* event);
+    virtual void SetParticlePosition(G4ThreeVector vec) override {};
```
ManuelHu, line 46:
```suggestion
    virtual void SetParticlePosition(G4ThreeVector) override {};
```

ManuelHu (approved):

PR #29, opened 2023-01-10
General build fixes 

drbenmorgan:
Building remage on macOS Ventura revealed a few issues which this PR aims to address:

- Consistent use of `PROJECT_{SOURCE,BINARY}_DIR` in place of the `CMAKE_...` versions. This is a "nice to have" just in case remage ever gets used as a subproject, and is a little clearer.
- Minor tidy of library CMakeLists.txt so all library commands are together, folllowed by those for executable, and finally install related ones. Again, purely for logic and clarity.
- Add the `FMT_HEADER_ONLY` preprocessing macro _everywhere_ in libremage, and for any target linking to it. On macOS at least, this is needed to prevent undefined symbol errors. Remage _does_ add this in `RMGLog.hh`, but not _everywhere_ `fmt` headers are pulled in.
  - This is a quick fix, longer term it's probably better to create/use a `fmt::fmt` imported target, maybe also hide `fmt` use entirely from the public interface of libremage.
- Fixed a couple of minor warnings for missing `override` on override virtual functions.
- _Very_ pedantic - it's "Geant4" rather than "GEANT4" (wearing by Geant4 🎩 😄 )!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
 endif()
 
 # include path for custom modules
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR}/cmake/modules/)
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/cmake/modules/)
 include(Colors)
 
 # get install directories names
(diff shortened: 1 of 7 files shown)


gipert:
Hi Ben, thanks a lot!

> Building remage on macOS Ventura revealed a few issues

We could add a macOS CI job, now that G4 official binaries are online?

> This is a quick fix, longer term it's probably better to create/use a fmt::fmt imported target, maybe also hide fmt use entirely from the public interface of libremage.

Hope we can actually switch to [C++20](https://en.cppreference.com/w/cpp/utility/format) at some point!

> Very pedantic - it's "Geant4" rather than "GEANT4"

I did not know! I assumed it had to be treated as an acronym ;) thanks

PR #30, opened 2023-01-16
Make EcoMug and BxDecay0 private dependencies

drbenmorgan:
Remage's event generators do not provide a public interface that exposes their internals such as EcoMug and BxDecay0. This makes these packages implementation details which a consumer of libremage does not need to be aware of. E.g. a consumer of libremage does _not_ need the EcoMug or BxDecay0 headers to be present, only the runtime library.

This PR uses a "Pimpl/compiler firewall" to hide the EcoMug and BxDecay0 instances from consumers. This does require that non-inline cons/destructors be used, and empty default member initialisation be used for the "Pimpl" `unique_ptr`. Whilst that may not appear elegant, it's a small price for hiding this stuff away (smaller install footprints, potentially faster code) - plus the _default_ constructor for `unique_ptr` is `nullptr` anyway, so largely moot.

### DIFF
--- a/include/RMGGeneratorCosmicMuons.hh
+++ b/include/RMGGeneratorCosmicMuons.hh
@@ -26,7 +26,7 @@
 #include "G4ParticleGun.hh"
 #include "G4VUserPrimaryGeneratorAction.hh"
 
-#include "EcoMug/EcoMug.h"
+class EcoMug;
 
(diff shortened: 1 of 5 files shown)


gipert:
Thanks a lot, I really did not pay attention to this aspect!

Issue #31, opened 2023-01-16
Status and recommended use of `clang-tidy`?


drbenmorgan:
In implementing #30, I tried out use of the current `.clang-tidy` file for remage and wanted to check/confirm its current status and recommended use in the project. Running it via the LLVM `run-clang-tidy` script on the remage code works fine, but there are a lot of issues reported, so I'm guessing it's not ready for full use yet?

Certainly, the list of checks is _probably_ too large right now. Not that we shouldn't have many checks, but probably need to build these up gradually, and from experience not all of them are particularly useful, or require tuning of parameters.

PR #32, opened 2023-01-23
Starting point/setup for .clang-tidy 

drbenmorgan:
As a follow up to #31, this demos a minimal `.clang-tidy` file with a small set of checks. These are _somewhat_ arbitrary, but generally work well and are safe to apply automatically (if that is desired). It's split into two commits to separate out:

1. The `.clang-tidy` change itself. This also includes a minor CMake update that copies the `.clang-tidy` file from source to build directory. LLVM's `run-clang-tidy` script is nominally supposed to find this in the source directory, but I've found this slightly flaky. Given that a model for operation is to use `run-clang-tidy` in the (or point it to the) build directory only, copying the config file here is generally reliable (to be reviewed though!).
2. The second commit shows what happens to the code after a `run-clang-tidy -fix -format` operation with the updated config file. Pretty minimal as expected, but shows it works!

This is of course just a demo, so fixes/updates etc very welcome. There are of course _plenty_ of additional checks that could be applied, but perhaps this is something to do step-by-step in future PRs. Some are "obvious" but may not always work nicely with the `-fix` optional without developer intervention.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,11 +1,19 @@
+---
+Checks: "-*, \
+  modernize-deprecated-headers, \
+  modernize-make-shared, \
+  modernize-make-unique, \
+  modernize-redundant-void-arg, \
+  modernize-use-auto, \
+  modernize-use-equals-default, \
+  modernize-use-nullptr, \
+  modernize-use-using, \
+  performance-faster-string-find, \
+  performance-for-range-copy, \
+  performance-inefficient-string-concatenation, \
+  performance-trivially-destructible, \
+  readability-delete-null-pointer, \
+  readability-duplicate-include, \
+  readability-string-compare"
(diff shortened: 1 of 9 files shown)


gipert:
Really nice! Is there a pre-commit hook for clang-tidy we can enable?

drbenmorgan:
There's this one: https://github.com/pocc/pre-commit-hooks, but trying it out didn't work too well:

1. It fails if clang-tidy isn't present (would have expected it to skip?)
2. It doesn't run properly from a build directory as it looks for the compilation database in the source directory.

There don't appear to be options to control that, so not super stable/usable off the bat, but in principle do-able.

gipert:
Ok, it definitely does not look like something we want in our hook collection. But let's keep an eye on this, we should have it at some point!

PR #33, opened 2023-01-23
Minimize install footprint/clashes and permit uninstall

drbenmorgan:
Just a small tune of what gets installed, and how:

- As remage currently installs all its headers direct in an `include/` subdirectory of `CMAKE_INSTALL_PREFIX`, rename `ProjectInfo.h` to `RMGConfig.h` to avoid clashes.
  - This also makes this header follow the convention of the other remage headers
- Exclude the CLI11 and EcoMug headers from the install as these are no longer needed by downstream consumers (since #30 was merged)
- Only link remage binaries to targets they directly use (CMake will handle transitive deps).
- Only log new or updated installs when running "make install" or similar. This helps make it clear exactly what happened.
- Error out if we try and install any file with an absolute path rather than one relative to `CMAKE_INSTALL_PREFIX`. Just a safeguard against install commands _not_  using the `CMAKE_INSTALL_....` paths.
- Provide an "uninstall" target per standard build expectations. Removes files, but not directories, which is deliberate (don't want to stomp on system/shared dirs).


gipert:
Wonderful, thanks!

PR #63, opened 2024-03-27
Add command /RMG/Geometry/RegisterDetector

ManuelHu:
* We need our own messenger for this, as `G4GenericMessenger` only supports binding to functions with <= 2 arguments.
* This also adds a safeguard against changing the list of active detectors after the detector has been constructed. Until now this silently failed/did nothing.
  This problem cannot be reached by the new command (it is bound to the `PreInit` state), but could be reached by custom C++ code.

### DIFF
--- a/include/RMGHardware.hh
+++ b/include/RMGHardware.hh
@@ -26,6 +26,7 @@
 #include "G4Region.hh"
 #include "G4VUserDetectorConstruction.hh"
 
+#include "RMGHardwareMessenger.hh"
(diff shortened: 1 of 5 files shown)


gipert:
Looks great, thanks Manuel.

ManuelHu:
Just a note: this still does not work, I am currently trying out how to fix it.

The analysis manager is set up before loading the macro file _at all_, so it is disables (and later crashes).

ManuelHu:
follow-up: #65

PR #64, opened 2024-04-01
cli: add option to set output file, type converter for log level

ManuelHu:
This is already possible in a macro file, but I would find it useful to be able to set this also with the cli parameters. (i.e. for GDML files, both ways are also valid)


ManuelHu:
Note that this is particularly ugly, but the ordering of the CLI11 include after the namespaced defintion of the converter function is necessary. Otherwise it would not be picked up by CLI11...

another note: Having three options to control log verbosity (`-v` to increase, `-q` to decrease, and `-l` to set it directly) seems like overkill to me.

PR #71, opened 2024-04-11
clang-tidy: add more modernize-* checks

### LINKED ISSUE #36: Run `clang-tidy` as pre-commit hook
There's this one: https://github.com/pocc/pre-commit-hooks, but trying it out didn't work too well:

* It fails if clang-tidy isn't present (would have expected it to skip?)
* It doesn't run properly from a build directory as it looks for the compilation database in the source directory.

There don't appear to be options to control that, so not super stable/usable off the bat.

ManuelHu:
I kept all checks that have a large volume of changes disabled.

fixes #36

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -8,9 +8,17 @@ Checks: "-*, \
   performance-for-range-copy, \
   performance-inefficient-string-concatenation, \
   performance-trivially-destructible, \
-  readability-delete-null-pointer, \
-  readability-duplicate-include, \
-  readability-string-compare"
+  readability-*, \
+  -readability-magic-numbers, \
+  -readability-convert-member-functions-to-static, \
+  -readability-isolate-declaration, \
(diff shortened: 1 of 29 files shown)


gipert:
Could you check again what is the situation with https://github.com/legend-exp/remage/issues/36?

ManuelHu:
I implemented a simple pre-commit hook for clang-tidy
* It does not fail if clang-tidy is not installed. Unfortunately there is no way to see this from the CLI output.
* It works with a separate build dir
* Unfortunately it "fails" if it applied any fixes, even when no error is remaining after that (that is just how the exit code of `clang-tidy` works...)


Currently there is no check that one used `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`  for cmake. Should we set this globally in the CMakeLists? Apart from a small (some 100 kB) storage footprint, I don't see any explicit downsides from it...

PR #77, opened 2024-04-25
Fix CI tests not failing, fix some other minor bugs

### LINKED ISSUE #76: CI does not fail if tests fail
1. CI tries to run tests with visualization
2. the runner has - apparently - no running X server
3. geant4 aborts before even trying to run the test, i.e. `***** Batch is interrupted!! *****`
4. but it is not setting any nonzero status code, so the tests are detected as passed... :-(

ManuelHu:
* Fix bug introduced in PR #73
* Introduce a new class `RMGExceptionHandler` to catch genat4 warning/errors, and to adjust the exist code accordingly
* For this to give us any reasonable results, I demoted some warnings to summary messages. Now you can run the tests without any warnings
    * It is not really a reason to warn if the run stats are inaccurate because of low event numbers
    * Also do not warn if no persistency is enabled, and no detectors are registered.
* Do not test anything with visualization in CI, this will not work, see #76
* This also fixes some wrong stats on case of aborted runs (and adds a warning for that)

fixes #76

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -25,4 +25,4 @@ CheckOptions:
   value: true
 
 FormatStyle: file
-HeaderFilterRegex: 'include/.*.hh'
+HeaderFilterRegex: 'include/.*\.(hh|icc)'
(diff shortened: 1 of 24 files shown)


ManuelHu:
still merging, all crashes are likely there for a long time - all confinement tests were never really run on CI

PR #87, opened 2024-05-02
Add hadronic physics

MoritzNeuberger:
The implementation is similar to MaGe's implementation. That means one has a boolean flag `fUseHadPhys` set via a macro command defining whether to use the hadronic physics. Then, the user can use the standard value or choose a different one from the options via another macro command.

Changes:
- [x] added an option to select different hadronic physics lists similar to the EM physics lists
- [x] added thermal neutron scattering physics
- [ ] add standard values recommended for ParticleHP (see commented part in `RMGPhysics.cc`) --> Have it hard coded? Have a mechanism such that the standard values are used unless the user specifies it? Maybe instead just set them in the environment variables and let the user control it that way.

FYI: I also found that the EM physics has a boolean `fUseLowEnergyEM` that is not currently used. It should also be implemented.


ManuelHu:
Would it make sense to add a `kNone` option (later also for the LowEnergyEM...), that would be essentially a no-op? Then we could not miss to use boolean flags again...

ManuelHu:
Looks good, just a small note if someone reads this: the environment variables for setting the ParticleHP parameters have been removed in g4 11.0.

So now, the corresponding (core) macro commands have to be used for this. With this, it also does not make sense to duplicate this functionality in the remage physics list messenger.

PR #113, opened 2024-08-10
disallow setting random engine in MT mode

ManuelHu:
also improve command guidance for engine selection


ManuelHu:
closed in favor of #114

PR #119, opened 2024-08-17
ci: check if command reference guide is up-to-date

ManuelHu:


gipert:
Also because of this, we should start writing some developer guide :)

PR #121, opened 2024-08-18
add optional isotope filter output scheme

### LINKED ISSUE #86: Output scheme option to write only when a certain isotope  is produced
Especially for cosmogenic production studies, a large amount of exposure has to be simulated. 
To avoid writing too much output, I propose a macro command that specifies that events are only stored when a specified isotope is produced.
It should be a relatively simple condition in the EndOfEventAction.

ManuelHu:
@MoritzNeuberger @EricMEsch is that what you wanted?

note: this is mostly untested

fixes #86 ?

### DIFF
--- a/docs/g4manual2rst.py
+++ b/docs/g4manual2rst.py
@@ -35,6 +35,7 @@ def remove_whitespace_lines_end(lines: list):
 
 idx = 0
 in_cmdblock = False
+in_guidance = False
 lastlevel = -1
 leveldiff = 0
 
@@ -52,9 +53,14 @@ for line in inlines:
         lastlevel = -1
         leveldiff = 0
     elif in_cmdblock and (line == "Guidance :"):
-        pass
-    elif in_cmdblock and (inlines[idx - 1] == "Guidance :") and not line.startswith(" " * 2):
-        outlines.extend([line, ""])
+        in_guidance = True
+    elif in_cmdblock and in_guidance and not line.startswith(" "):
+        if line.startswith("note: "):
+            outlines.extend([".. note ::", "", (" " * 4) + line[6:], ""])
+        elif line.strip() == "":
+            in_guidance = False
+        elif line != "":
+            outlines.extend([line, ""])
     elif in_cmdblock and line == " Commands : " and not inlines[idx + 1].startswith(" " * 3):
         # ignore directories with no commands.
         pass
@@ -62,6 +68,8 @@ for line in inlines:
         # ignore directories with no commands.
         pass
     elif in_cmdblock and line != "":
+        in_guidance = False
+
         stripped_line = line.lstrip()
         indent = math.ceil((len(line) - len(stripped_line)) / 2)
         if line.startswith(" Range of parameters :"):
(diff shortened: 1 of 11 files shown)


MoritzNeuberger:
Uhhh, that looks pretty nice! I love it.

EricMEsch:
Looks good! I will try to do some more extensive tests on it!

EricMEsch:
I have found an issue that arises from the way the StackingActionNewStage is handled:
```cpp
std::optional<bool> RMGGermaniumOutputScheme::StackingActionNewStage(const int stage) {
  if (!fDiscardPhotonsIfNoGermaniumEdep) return true;
  auto run_man = RMGManager::Instance()->GetG4RunManager();
  const auto event = run_man->GetCurrentEvent();
  // discard all waiting events, as there was no energy deposition in Germanium.
  return !ShouldDiscardEvent(event);
}
```
It returns the request only with `std::nullopt` when `stage != 0`. This means for the main stage `0`, as long as this outputscheme is set, it will either request a new stage or request there to be no new stage. (especially if the macro `fDiscardPhotonsIfNoGermaniumEdep` is turned off, it will still request for a new stage).

The issue is the StackingAction asks all of the outputschemes whether they want a new stage or not and gives an error on conflicting requests. This means as soon as the Germanium and IsotopeFilter Outputschemes are registered, there will always be a conflict, if not both are set active and by pure chance happen to agree.
```cpp
for (auto& el : fRunAction->GetAllOutputDataFields()) {
    auto request_stage = el->StackingActionNewStage(fStage);
    if (!request_stage.has_value()) continue; // this output scheme does not care.

    if (!should_do_stage.has_value() || should_do_stage.value() == request_stage.value()) {
      should_do_stage = request_stage;
    } else {
      RMGLog::Out(RMGLog::error,
          "Conflicting requests for new stage termination in RMGStackingAction.");
    }
  }
```

This should probably be fixed by changing to `if (!fDiscardPhotonsIfNoGermaniumEdep) return std::nullopt;` (and similar in IsotopeFilterOutputscheme).
That way conflicts can only arise if both macros are set.

EricMEsch:
Yes that seems like a reasonable solution. In that case there could also be made an argument to have `true` take precedence over `false` when a conflict arises, in order to give user initialized outputschemes the option to force generate optical photons when a given condition is met? As all of the integrated Outputschemes would either return `false` or `std::nullopt`

EricMEsch:
I have also found the issue that the features are not thread-safe. 

Steps to produce crash:
add lines:
```
/RMG/Output/ActivateOutputScheme IsotopeFilter
(/run/initialize)
/RMG/Output/IsotopeFilter/AddIsotope 77 32
/RMG/Output/IsotopeFilter/DiscardPhotonsIfIsotopeNotProduced true
```
to example 05-MUSUN (i dont think the example matters, but this uses 2 threads as default)
result: Crashes in RMGIsotopeFilterOutputScheme line 52 `auto info = event->GetUserInformation();`
due to segmentation fault.

After some digging i am pretty certain it crashes because `run_man->GetCurrentEvent()` in `RMGIsotopeFilterOutputScheme::StackingActionNewStage(...)` is not thread-safe and returns `null` in multithread mode? I was unable to exactly find which runmanager is created by the RunManagerFactory but i guess the standard G4MTRunManager?

The same issue also occurs in the Germanium Output System, so i am not sure if i should open up a new issue, as this is an error that is not being introduced by this PR, but rather just being reproduced.

ManuelHu:
Thanks - I think no-one ever actually used filtering with MT mode on. You can open an issue if you want to, but I will fix it anyway :-)
Just fyi, by default, G4 11.0+ does not use `G4MTRunManager` but `G4TaskingRunManager`...

PR #122, opened 2024-08-18
docs: add simple devguide

ManuelHu:
@gipert this is just a copy of legend-pydataobj's devguide, shortened and adjusted for remage

I also added a section on automatic docs generation

### DIFF
--- a/docs/developer.rst
+++ b/docs/developer.rst
@@ -1,0 +1,128 @@
+Developer's guide
+=================
+
+The following rules and conventions have been established for the package
+development and are enforced throughout the entire code base. Merge requests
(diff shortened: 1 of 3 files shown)


docs/developer.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Installing dependencies
+-----------------------
+
+.. include:: _dependencies.rst
+
+Building
+--------
+
+**remage** currently only supports an out-of-tree buiold using cmake. To build and
```
gipert, line 30:
```suggestion
**remage** currently only supports an out-of-tree buil using CMake. To build and
```

PR #123, opened 2024-09-02
Add Grabmayrs gamma-cascade-reader and a custom neutron capture process 

EricMEsch:
The first commit is intended to add a refurbished version of the gamma-cascade-reader originally written by @MoritzNeuberger for WLGD. It employs a custom `RMGNeutronCaptureProcess` inheriting from `G4HadronicProcess` and replacing the native `G4NeutronCaptureProcess`. This process is supposed to employ the gamma cascades from Peter Grabmayr (https://link.springer.com/article/10.1140/epjc/s10052-023-11602-y) upon neutron capture on specified Isotopes, otherwise keeping the standard process. It manually employs the cross section datasets that are usually employed with `Shielding` and takes over all of the interactions from the native `G4NeutronCaptureProcess`. Works with 3 of the 4 physics list, as `QGSP_BERT_HP` does employ a different process for "nCapture" (although this could be fixed with some investigation, for now this will result in just the native process being used and an error message thats not crashing the application, if the `UseGrabmayrsGammaCascades` macro is set).

There are certainly other options for this feature, but overall to me this seems like a "smoother" solution, than doing it in the tracking action like in WLGD. However, if there are other ideas that suit the concept of ReMaGe more, i am open ears. This is just a proposal.

The second commit provides an example to test this new feature, while employing a user-defined optional output scheme to write out all of the isotopes created through `nCapture` or `RMGnCapture`. It provides two macros to compare the native result to this new custom version.

The example does not need to be pulled, but it allows for quick tests (and shows how a user-defined optional output scheme would work)


ManuelHu:
Did you test this with MT mode on? I do not suppose that all the file-reading stuff is thread-safe, or is it? (making the reader instance thread-local should be enough. It has to be constructed on each thread anyway, otherwise the registered commands will not really work in some edge cases.)

EricMEsch:
I did test it, thats why in the UserAction the instance is called in the `Build()` and `BuildForMaster()` as it seems in Single thread mode it requires it to be in `Build()` and in the multi threaded case it seems to work with the instance first being called in `BuildForMaster()`.

This worked for my tests, but i have to admit i am not too sure if that solution works for all edge cases, as i am not entirely sure about the interaction of singletons in MT mode.

ManuelHu:
I do not have time to test it now, but I would guess that the current code will fail if you move `/RMG/GrabmayrGammaCascades/SetGammaCascadeFile` commands after `/run/initialize`.

I would be more concerned with only using one global instance without explicit locking, when it comes to file reading. Afaik, the C++ standard does not guarantee thread-safety around stream objects (see section 30.2.3 in the [C++ 17 draft](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf))

> I did test it, thats why in the UserAction the instance is called in the Build() and BuildForMaster() as it seems in Single thread mode it requires it to be in Build() and in the multi threaded case it seems to work with the instance first being called in BuildForMaster().

I think, `BuildForMaster()` will not be called in single-threaded mode; only `Build()` is called...

The initialization will be mostly fine, even though _technically_ UB, I suppose.

src/RMGGrabmayrGCReader.cc
```diff
@@ -0,0 +1,161 @@
+#include "RMGGrabmayrGCReader.hh"
+
+#include "G4Tokenizer.hh"
+#include "Randomize.hh"
+
+#include "RMGLog.hh"
+
+
+RMGGrabmayrGCReader* RMGGrabmayrGCReader::instance = nullptr;
```
ManuelHu, line 9:
```suggestion
G4ThreadLocal RMGGrabmayrGCReader* RMGGrabmayrGCReader::instance = nullptr;
```

include/RMGGrabmayrGCReader.hh
```diff
@@ -0,0 +1,69 @@
...
+    G4int m;               // multiplicity of gamma cascade
+    G4int em;              // missing energy [keV]
+    std::vector<G4int> eg; // energies of photons [keV]
+};
+
+
+class RMGGrabmayrGCReader {
+  public:
+
+    static RMGGrabmayrGCReader* GetInstance();
```
ManuelHu, line 41:
```suggestion
    static G4ThreadLocal RMGGrabmayrGCReader* GetInstance();
```

src/RMGUserAction.cc
```diff
@@ -70,6 +72,8 @@ void RMGUserAction::Build() const {
   this->SetUserAction(new RMGStackingAction(run_action));
   this->SetUserAction(stepping_action);
   this->SetUserAction(tracking_action);
+  // Maybe add a check for multithreaded and only do this if not?
```
ManuelHu, line 75:
```suggestion
```

ManuelHu (commented):
That _should_ work (not compile-tested though).

EricMEsch:
This fail seems to be totally unrelated to the changes commited? (During runtime all of the added code is inactive if the macro is not set)

ManuelHu:
> This fail seems to be totally unrelated to the changes commited? (During runtime all of the added code is inactive if the macro is not set)

Yes, it's just the typical test being flaky, as G4 **11.0** has apparently some navigation bugs :-)

src/RMGGrabmayrGCReader.cc
```diff
@@ -0,0 +1,161 @@
...
+    RMGLog::Out(RMGLog::detail, "Random start location: ", start_location);
+    for (int j = 0; j < start_location; j++) std::getline(*file, line);
+  }
+  return std::move(file);
+}
+
+void RMGGrabmayrGCReader::SetGammaCascadeFile(const G4String& params) {
+  // Convert token to Z, A and path/to/file
+  G4Tokenizer tokenizer(params);
+  G4String zStr = tokenizer(",");
```
ManuelHu, line 108: I do not really like this, but I known that it is the best one can do with the `G4GenericMessenger` (because it - completely arbitrarily(*) - limits the argument count to 2.)
Otherwise one could implement a full custom messenger class just for this command, see `RMGHardwareMessenger`...

(*)small rant: There is not even a good technical reason for that, most of the code already can handle >2 arguments just fine, they did just not define their template types for it.

src/RMGGrabmayrGCReader.cc
```diff
@@ -0,0 +1,161 @@
...
+  RMGLog::Out(RMGLog::detail, "Opening file ", file_name);
+  std::unique_ptr<std::ifstream> file = std::make_unique<std::ifstream>(file_name);
+
+  if (z == 0 || a == 0)
+    RMGLog::Out(RMGLog::fatal, "Isotope Z:" + std::to_string(z) + " A: " + std::to_string(a) +
+                                   " does not exist. Are you sure you used the correct format?");
+  if (!file || !file->is_open())
+    RMGLog::Out(RMGLog::fatal, "Gamma cascade file: " + file_name + " not found! Exit.");
+
+  file = SetStartLocation(std::move(file));
```
ManuelHu, line 123:
```suggestion
  SetStartLocation(*file);
```

src/RMGGrabmayrGCReader.cc
```diff
@@ -0,0 +1,161 @@
...
+
+  fCascadeFiles.insert({{z, a}, std::move(file)});
+}
+
+void RMGGrabmayrGCReader::RandomizeFiles() {
+  RMGLog::Out(RMGLog::detail, "(Un)-Randomizing start locations");
+  for (auto& el : fCascadeFiles) { el.second = SetStartLocation(std::move(el.second)); }
```
ManuelHu, line 130:
```suggestion
  for (auto& el : fCascadeFiles) { SetStartLocation(*el.second); }
```

src/RMGGrabmayrGCReader.cc
```diff
@@ -0,0 +1,161 @@
...
+
+    file->clear();                 // clear EOF flag
+    file->seekg(0, std::ios::beg); // move to beginning of file
+
+    int start_location = (int)(n_entries_in_file * G4UniformRand() + header_length);
+
+    RMGLog::Out(RMGLog::detail, "Random start location: ", start_location);
+    for (int j = 0; j < start_location; j++) std::getline(*file, line);
+  }
+  return std::move(file);
```
ManuelHu, line 102: I think you should be able to use a reference instead of a unique_ptr here, to avoid the repetitive std::move-ing around (not compile tested :-)

```suggestion
void RMGGrabmayrGCReader::SetStartLocation(std::ifstream& file) {
  if (!file.is_open())
    RMGLog::Out(RMGLog::fatal, "The file is not open to set start location! Exit.");
  file.clear();                 // clear EOF flag
  file.seekg(0, std::ios::beg); // move to beginning of file
  // Skip Header
  std::string line;
  int header_length = 0;
  do {
    std::getline(file, line);
    header_length++;
  } while (line[0] == '%' || (line.find("version") != std::string::npos));

  // In case the Random start location macro is set
  if (fGammaCascadeRandomStartLocation) {
    int n_entries_in_file = 0;
    // Seems excessiv to read through the entire file, there has to be a quicker way?
    while (std::getline(file, line)) n_entries_in_file++;

    file.clear();                 // clear EOF flag
    file.seekg(0, std::ios::beg); // move to beginning of file

    int start_location = (int)(n_entries_in_file * G4UniformRand() + header_length);

    RMGLog::Out(RMGLog::detail, "Random start location: ", start_location);
    for (int j = 0; j < start_location; j++) std::getline(file, line);
  }
```

src/RMGGrabmayrGCReader.cc
ManuelHu: It should be possible to use as nested class https://en.cppreference.com/w/cpp/language/nested_types? Then you would not need to add additional files...

src/RMGGrabmayrGCReader.cc
```diff
@@ -0,0 +1,183 @@
...
+}
+
+void RMGGrabmayrGCReader::SetGammaCascadeFile(const G4int z, const G4int a, const G4String file_name) {
+
+  RMGLog::Out(RMGLog::detail, "Opening file ", file_name);
+  std::unique_ptr<std::ifstream> file = std::make_unique<std::ifstream>(file_name);
+
+  if (z == 0 || a == 0)
+    RMGLog::Out(RMGLog::fatal,
+        "Isotope Z:" + std::to_string(z) + " A: " + std::to_string(a) + " does not exist.");
```
ManuelHu, line 109: this can still be simplified:
```suggestion
    RMGLog::OutFormat(RMGLog::fatal, "Isotope Z: {} A: {} does not exist.", z, a);
```

src/RMGGrabmayrGCReader.cc
```diff
@@ -0,0 +1,183 @@
...
+      .SetStates(G4State_PreInit, G4State_Idle);
+
+  // SetGammaCascadeFile cannot be defined with the G4GenericMessenger (it has to many parameters).
+  fUIMessenger = std::make_unique<GCMessenger>(this);
+}
+
+
+RMGGrabmayrGCReader::GCMessenger::GCMessenger(RMGGrabmayrGCReader* reader) : fReader(reader) {
+  fGammaFileCmd = new G4UIcommand("/RMG/GrabmayrGammaCascades/SetGammaCascadeFile", this);
+  fGammaFileCmd->SetGuidance("Set the Z, A and /path/to/file for the gamma cascade employed upon "
```
ManuelHu, line 151: I do not really like the message here. It seems to technical for a general description of what this command does. Maybe something akin "Set a gamma cascade file to be used upon neutron capture on a specifc isotope"?
(The parameters are already described by their own docstring, thats fine.)

src/RMGPhysics.cc
```diff
@@ -222,6 +225,34 @@ void RMGPhysics::ConstructProcess() {
...
 
+    if (fUseGrabmayrGammaCascades) {
+      // Apply RMG custom neutron capture
+      // Mostly similar to examples/extended/Hadr04
+      G4ProcessManager* pManager = G4Neutron::Neutron()->GetProcessManager();
+      G4ProcessVector* processVector = pManager->GetProcessList();
+      G4NeutronCaptureProcess* neutronCaptureProcess = nullptr;
+      // Find the existing neutron capture process
+      neutronCaptureProcess =
+          dynamic_cast<G4NeutronCaptureProcess*>(pManager->GetProcess("nCapture"));
```
ManuelHu, line 236:
```suggestion
      auto pManager = G4Neutron::Neutron()->GetProcessManager();
      auto processVector = pManager->GetProcessList();
      // Find the existing neutron capture process
      auto neutronCaptureProcess =
          dynamic_cast<G4NeutronCaptureProcess*>(pManager->GetProcess("nCapture"));
```

src/RMGPhysics.cc
```diff
@@ -222,6 +225,34 @@ void RMGPhysics::ConstructProcess() {
...
+
+      // Overwrite the old Process, keeping all of the interactions
+      if (neutronCaptureProcess) {
+        RMGLog::Out(RMGLog::detail, "Overwriting NeutronCaptureProcess");
+        pManager->RemoveProcess(neutronCaptureProcess);
+        RMGNeutronCaptureProcess* RMGProcess = new RMGNeutronCaptureProcess();
+        // HP cross section data set not naturally in G4NeutronCaptureProcess
+        G4ParticleHPCaptureData* dataSet = new G4ParticleHPCaptureData();
```
ManuelHu, line 244:
```suggestion
        auto RMGProcess = new RMGNeutronCaptureProcess();
        // HP cross section data set not naturally in G4NeutronCaptureProcess
        auto dataSet = new G4ParticleHPCaptureData();
```

ManuelHu:
Now it really looks better. I still have some minor (mostly cosmetci) remarks, then it should be good from my side :-)

src/RMGNeutronCaptureProcess.cc
```diff
@@ -0,0 +1,124 @@
...
+  theTotalResult->ProposeWeight(fWeight);
+
+  // Can skip all of the Checks for illegal track status, as in that case no Applicable isotope would have been produced!
+
+  // Do our Physics.
+  GammaCascadeLine input = CascadeReader->GetNextEntry(z, a);
+  G4ThreeVector location = aTrack.GetPosition();
+  G4double time = aTrack.GetGlobalTime();
+
+  // Expect no deponition as the total energy will be distributed
```
ManuelHu, line 94:
```suggestion
  // Expect no deposition as the total energy will be distributed
```

src/RMGGrabmayrGCReader.cc
```diff
@@ -0,0 +1,183 @@
...
+      ->DeclareMethod("SetGammaCascadeRandomStartLocation",
+          &RMGGrabmayrGCReader::SetGammaCascadeRandomStartLocation)
+      .SetGuidance("Set the whether the start location in the gamma cascade file is random or not")
+      .SetGuidance("0 = don't")
+      .SetGuidance("1 = do")
+      .SetCandidates("0 1")
+      .SetDefaultValue("0")
+      .SetStates(G4State_PreInit, G4State_Idle);
+
+  // SetGammaCascadeFile cannot be defined with the G4GenericMessenger (it has to many parameters).
```
ManuelHu, line 144:
```suggestion
  // SetGammaCascadeFile cannot be defined with the G4GenericMessenger (it has too many parameters).
```

PR #125, opened 2024-09-21
make clang-tidy happy with RMGnCapture

ManuelHu:
/cc @EricMEsch 

this should not change any functionality, and just fixes complaints by clang-tidy. Mostly:

* `const` specifiers missing or ignored
* non-matching function parameter names between declaration and definition

### DIFF
--- a/include/RMGGrabmayrGCReader.hh
+++ b/include/RMGGrabmayrGCReader.hh
@@ -38,11 +38,11 @@ struct GammaCascadeLine {
 class RMGGrabmayrGCReader {
   public:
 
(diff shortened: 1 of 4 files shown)


EricMEsch:
Sounds good, so this means in case of any future PRs i should also run clang-tidy seperately? I thought it was part of the pre-commit hook

ManuelHu:
It is part of the pre-commit hook, but only if you install it manually beforehand. (i.e. the `clang-tidy` package for ubuntu). It is not installed in the `remage-base` container.

PR #132, opened 2024-10-17
Only delay OpticalPhotons if discard macro is set

EricMEsch:
Currently, the optical photons will always be put on the waiting stack, even if the discard macro is not set. This means, they will be simulated anyways, but will consume (a lot of) memory on the stack while they are waiting.

This will only put them on the stack if the possibility exists for them to be discarded later.


src/RMGGermaniumOutputScheme.cc
```diff
@@ -168,7 +168,9 @@ std::optional<G4ClassificationOfNewTrack> RMGGermaniumOutputScheme::StackingActi
   if (stage != 0) return std::nullopt;
 
   // defer tracking of optical photons, irrespective of our settings.
```
ManuelHu, line 170:
```suggestion
  // defer tracking of optical photons.
```

Issue #135, opened 2024-10-22
Times for decay chains / killed nuclei


tdixon97:
* If we simulate decay chains with steps with very long livetimes we may get numerical issues when time-windowing the outputs. Since we look at the time-difference between G4Steps and the first step in the evt, if these are very large the difference will be 0.
* I simulated Ra-224 with the following commands.

```/RMG/Generator/Select GPS  
/gps/particle ion
/gps/energy 0 eV
/gps/ion 88 224
/process/had/rdm/nucleusLimits 208 224 81 88
```
* This leads to times that are large (as expected due to few days Ra-224 half-life), eg times for one event are
 ```[3.11e+14, 3.11e+14, 3.11e+14, 3.11e+14, ..., 3.11e+14, 3.11e+14, 3.11e+14] ns``` 
 but these still allow to determine delta t with < 1 ns precision.

* But if we would like to simulate decays with longer half-lives (Ra-226, U-238 etc.) we get into issues, as well as if a part of the decay chain has a long half-life.
* However, when I tried simulating Ra-226 I got no events in the output file, so the nuclei is being killed before the decay (without a warning message etc.). Seems related to [stepping-action](https://github.com/legend-exp/remage/blob/main/src/RMGSteppingAction.cc#L30), maybe we need some warnings if the user tries to simulate decays with T12 ~ time to kill nuclei?
* @ManuelHu  do we need a solution to allow simulation of arbitary decays including those with long half-lives?

ManuelHu:
Geant4 11.2 changed the default value of the max lifetime to 1 year (yeah, I also think this is bad to change it in a minor release)

reset to old default (maybe we should add this to remage to make it behave consistently):
```
/process/had/rdm/thresholdForVeryLongDecayTime 1.0e+60 year
```

> Seems related to [stepping-action](https://github.com/legend-exp/remage/blob/main/src/RMGSteppingAction.cc#L30),

this feature to kill long-lived daughter nuclei should be disabled by default; `fDaughterKillLifetime` should be -1, if you do not set it in a macro.

tdixon97:
Ok that makes sense.
So we shouldn't have any issues with numerical precision for times < 1 yr (a double can hold the number of ns in a year without any loss of precision below 1ns).
Although if we increase this using the command you we will start to have the issues I was worried about regarding numerical issues with time-differences for windowing into events.
Should we try to think of a solution? Maybe not letting users simulate these decays and exiting with an error is the best short-term solution

[...]

PR #136, opened 2024-10-22
Add option to reset global time after initial decay

ManuelHu:
see #135

### DIFF
--- a/docs/rmg-commands.md
+++ b/docs/rmg-commands.md
@@ -502,6 +502,7 @@ Commands for controlling physics processes
 ### Commands
 
 * `DaughterNucleusMaxLifetime` – Determines which unstable daughter nuclei will be killed, if they are at rest, depending on their lifetime.
+* `ResetInitialDecayTime` – If the initial step is a radioactive decay, reset the global time of all its secondary tracks to 0.
 
 ### `/RMG/Processes/Stepping/DaughterNucleusMaxLifetime`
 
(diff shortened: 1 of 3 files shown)


ManuelHu:
@tdixon97 I also added a warning for large global times after ~285 years of simulation time. This corresponds to 1 us time resolution.

```c++
std::cout << std::nextafter(285 * CLHEP::year, INFINITY) - 285 * CLHEP::year) / CLHEP::us << std::endl;
// -> 1.024
```

the warning is only shown once per thread to avoid it being shown for each event.

tdixon97:
I think the warning is good but then I wonder if we want the limit by default to be set to this?

ManuelHu:
that would give us non-deterministic, inconsistent and for users even more confusing behaviour: the limit does affect the true (sampled) lifetime, and not the half-life. So some decays would go through, and some not.
also it applies to the time of the individual decays,  and not the (summed) global time in the event. So 3 decays with 100 years lifetime would not trigger the threshold, but only this warning.

Also I would suppose that - when I want to simulate some specific decay - remage should also just do how it is instructed.

tdixon97:
Hmm I am still not sure, I think `remage` should perform some sanity checks and make the user force it to do something that would give a nonsense output. What do you think @gipert ? Maybe the check should be on the half-lives not on the sampled half-lives as you say

gipert:
I think the check should be on half-lives, not the sampled decay times. Will anyone really want to cut on the sampled times?

tdixon97:
I agree, cutting on sampled times is not a nice behaviour really

ManuelHu:
probably not. This is why this PR is raising that builtin limit acting on sampled lifetimes. But we cannot really disable this functionality. (also because the API to set the limit from C++ only exists in 11.2+, and not in 11.0-11.1).

But this is a more complex issue, as Toby wants to enforce a maximum time uncertainty in output.
This cannot achieved by cutting on individual decay half-lifes. It would have to be a **check on the sum of all half-lives in a decay chain**. But this information is not readily available in Geant4, so we would need to track it on our own.

Such a check would give us reproducability, but would not really solve the output timing issue? If a chain approaches this 285yr region, a lot of events would still have time uncertainty > 1us _even after that cut_...
Example: Ar39 has 268yr half-life...

---

or we just want an option to constrain half-lifes of daughter nuclei (not summed); we already have this option. I think that having options for primaries do not make sense at all - if a user requests a specific isotope, they should already know best that they really want it...

PR #140, opened 2024-11-02
Mention pre-built binary on Docker Hub

gipert:


ManuelHu:
should we switch to `apptainer pull` (instead of `pull) here? As we have seen in both our local tests, that would have less problems in some cases (i.e. old apptainer versions)?

https://apptainer.org/docs/user/main/cli/apptainer_pull.html

For me, it appears that `build` is intended to be used in a larger setup to create writable images for later modifications. That it can also pull images to a read-only sif image is not its intended use, I suppose.

gipert:
Ah, I wasn't aware of this. Does it use less resources? If yes we should recommend it.

ManuelHu:
I read some more, and both should be equivalent. The docs even say that `pull` uses `build` under the hood for docker images.

It's just the docs that always use `pull` with docker images, but it should not matter.

PR #141, opened 2024-11-04
remage-cli: handle SIGUSR1 signal to preempt the current run

ManuelHu:
this could be useful to run it in preemptible QOS on some HPC clusters.


gipert:
ah yes very good idea :)

PR #142, opened 2024-11-04
add option to output float32 instead of float64 for some fields

ManuelHu:
Different simulations will have different needs (i.e. energy depositions in scintillators might not be that well localized than HPGe hits), so make it configurable per output scheme.

But it still feels ugly to have that many options....


src/RMGVertexOutputScheme.cc
```diff
@@ -118,6 +125,14 @@ void RMGVertexOutputScheme::DefineCommands() {
   fMessenger->DeclareProperty("SkipPrimaryVertexOutput", fSkipPrimaryVertexOutput)
       .SetGuidance("Do not store vertex/primary particle data.")
       .SetStates(G4State_Idle);
+
+  fMessenger->DeclareProperty("StoreLessPrecisePosition", fStoreLessPrecisePosition)
```
gipert, line 129: I would suggest something more specific than "less precise", maybe directly say "single precision" or so

PR #144, opened 2024-11-05
Add Dockerfile for automated Docker Hub builds

gipert:


docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
+ARG REMAGE_BASE_FLAVOR="latest"
+
+FROM gipert/remage-base:$REMAGE_BASE_FLAVOR
+
+LABEL maintainer.name="Luigi Pertoldi"
+LABEL maintainer.email="gipert@pm.me"
+
+ARG REMAGE_VERSION="v0.2.0"
```
ManuelHu, line 8: should we leave the fixed verison in here?

gipert: It should be overridden by the build hook.

PR #145, opened 2024-11-05
Update documentation

gipert:


ManuelHu (approved):

gipert:
I started with the tutorial (but want to add much more in the next days), please have a look.

@ManuelHu  @tdixon97

docs/tutorial.md
```diff
@@ -0,0 +1,202 @@
...
+
+![Geometry visualization](img/tutorial-pyg4-view.jpg)
+
+By following instructions in the <project:./install.rst> section, you should
+have access to the `remage` executable. We are now ready to simulate some
+particle physics with it.
+
+Like any other Geant4-based application, we need to configure the simulation
+with a macro file. Standard Geant4 commands as well as custom commands (see the
+<project:./rmg-commands.rst>) are available.
```
ManuelHu, line 118: those `project:./` links do not work (at least on RTD, did not try locally)

docs/tutorial.md
```diff
@@ -0,0 +1,202 @@
...
+/gps/energy 1000 keV
+
+/run/beamOn 50
+```
+
+We can finally pass the GDML geometry and the macro to the `remage` executable
+and look at the result!
+
+```console
+$ remage --interactive --gdml-files geometry.gdml vis-gammas.mac
```
ManuelHu, line 179: maybe add the option to select an output file? (also below in the note block). Or do you plan to have this in the next (not yet written) part?

docs/tutorial.md
```diff
@@ -0,0 +1,202 @@
...
+
+Like any other Geant4-based application, we need to configure the simulation
+with a macro file. Standard Geant4 commands as well as custom commands (see the
+<project:./rmg-commands.rst>) are available.
+
+At the beginning of the file, we have to register the "sensitive detectors" (in
+our simple case, the two HPGes and the LAr). *remage* offers several types of
+predefined detectors, targeting different physical quantities of the particles
+that interact with them. HPGes are of type `Germanium`, while the LAr is of
+type `scintillator`. Their difference in terms of simulation output will be
```
ManuelHu, line 124:
```suggestion
type `Scintillator`. Their difference in terms of simulation output will be
```

docs/tutorial.md
gipert: uff works locally...

gipert:
I think this looks better now. Anything else crucial I should include?

ManuelHu:
Maybe we should use a less confusing (i.e. more useful in the field of particle detectors) material for the source volume?

Apart from that it already looks very good :-)

tdixon97:
I like the source material

gipert:
Trying to fix #149 but it's a bit flaky, it does not recognize parameter description fields yet...

ManuelHu:
maybe we should move this to a seperate PR?

PR #151, opened 2024-11-07
Fix command docs

ManuelHu:
part of #145 moved out for easier code changes


gipert:
Seems like the admonitions and the description line after "Parameters :" are broken

ManuelHu:
admonitions yes, but which description line do you mean?

gipert:
https://remage--151.org.readthedocs.build/en/151/rmg-commands.html#rmg-geometry-registerdetector

The line after the "Parameter" bullet point should not start with a bullet

Issue #158, opened 2024-11-11
Add support for GRDM generator


gipert:
https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/Fundamentals/biasing.html?highlight=grdm#radioactive-decay-biasing

gipert:
In MaGe, this is used to simulate decay chain segments by setting start and stop nucleus.

```
/MG/generator/select RDMiso
/gun/energy 0 eV
/grdm/ion 228 89 0
/grdm/nucleusLimits 228 228 88 89
```
 This is of course analogous to GPS ion + livetime limit but is sometimes more "readable" in a macro.

ManuelHu:
it's 1 year; so much to short for most of our isotopes(*). Maybe we should set it in remage... (see also #135 where Toby noticed this)

(*) `thresholdForVeryLongDecayTime` applies both to primaries and secondaries (unlike our limit), so you might not even get any events simulated.

PR #160, opened 2024-11-12
physics: increase default time threshold for decays

ManuelHu:
as it was before geant4 11.2...

This avoids confusion and unforseen breakage. Also the higher threshold makes sense for most of our simulations?

what do you think @gipert @tdixon97 ?

see also #135 #158


tdixon97:
I think its a good idea but we should probably work out what time difference can be resolvable by double with the necessarily precision (idk 1 us). I think this is probably around 30,000 yrs (ie roughly 17 digits for double).
Then we should give an error or warning if the user requests a simulation with an intermediate step having dt > this.

ManuelHu:
the old default was actually 1e27 ns (~twice the age of the universe), and not 1e60 years.

Issue #153, opened 2024-11-10
Add option to directly use sensitive detector names to name output tables


gipert:
This would simplify detector identification in post-processing (when e.g. querying metadata). An exception should be thrown if the names are not all unique and if the user requests to have all data in one big table.

ManuelHu:
For the table mode per detector type we would still need to use the uids, but we could add a mapping table

gipert:
> For the table mode per detector type we would still need to use the uids

I don't see why other than having to change the code a bit...

gipert:
Oh I understand. Mabye we can just get rid of it?

ManuelHu:
This "flat" output mode is exactly what moritz requested in #85 , so probably not.

gipert:
Sorry, I misunderstood again. In the multi-table output mode we don't store a column with the UID, so it would not use more disk space. We could change the code to generate a random UID to be used internally and then name the table in the output after the sensitive volume name.

As I said at the beginning, the situation is different for the single-table mode and we should not implement what I'm proposing there.

ManuelHu:
> Maybe storing a string is not actually not bad since it should compress very well

 I did a quick test and stored _the same_ 100 byte long constant string for each vertex, with HDF5 output. The file size difference was ~ 72 MB (for 60 MB of additional string data, so it did not compress _at all_, it is actually even worse!). Also the runtime was increaed by a factor of 8.5.
So no, storing strings for every simulation hit will not be good :-(

tdixon97:
Ok, I think we should not move to a single flat table since writing a column thats another ID (not meaning anything) is confusing and processing the flat table is not easier. It should be fairly easy to combine in post-proccesing the table so Id stick with @gipert suggestion...

PR #161, opened 2024-11-12
output: use detector names as ntuple name

ManuelHu:
see #153 for discussion

* not yet an option (but an option would be easy to add)
* no solution for the flat output mode yet


ManuelHu:
@gipert now the option is added, should be ready

src/RMGManager.cc
```diff
@@ -376,18 +376,24 @@ void RMGManager::DefineCommands() {
   fOutputMessenger->DeclareProperty("NtuplePerDetector", fOutputNtuplePerDetector)
       .SetGuidance("Create a ntuple for each sensitive detector to store hits. Otherwise, store "
                    "all hits of one detector type in one ntuple.")
-      .SetParameterName("tree_per_det", false)
+      .SetParameterName("nt_per_det", false)
+      .SetStates(G4State_PreInit, G4State_Idle);
+
+  fOutputMessenger->DeclareProperty("NtupleUseVolumeName", fOutputNtupleUseVolumeName)
+      .SetGuidance("Use the sensitive volume name to name output ntuples.")
+      .SetGuidance("note: this only works if `NtuplePerDetector` is set to false.")
```
gipert, line 384: set to true?

ManuelHu: yes

gipert:
Looks good to me.

PR #165, opened 2024-11-21
Adjust examples to the new detector handling

EricMEsch:
The recent updates that moved the 
```c++
    enum DetectorType {
      kGermanium,
      kOptical,
      kScintillator,
    };
```
Out of `RMGHardware` broke the examples and the custom written c++ code, as the enum is now no longer part of the `RMGHardware` class. This went past the tests, as there is no extra compiled code in the tests.

I am not sure if the rest of the way the registration should be handled is still the same, but this works. Compile tested but did not check the output.

### DIFF
--- a/examples/02-hpge/main.cc
+++ b/examples/02-hpge/main.cc
@@ -8,10 +8,10 @@ int main(int argc, char** argv) {
   RMGManager manager("02-hpge", argc, argv);
   manager.SetUserInit(new HPGeTestStand());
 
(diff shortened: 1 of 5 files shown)


ManuelHu:
> I am not sure if the rest of the way the registration should be handled is still the same, but this works. Compile tested but did not check the output.

should have stayed the same. The naming has only been changed to break a circular dependency... (and the "typical" way to break that is impossible: one cannot forward declare nested structs/enums; thanks C++)

PR #168, opened 2024-11-27
remove some more unwanted output

ManuelHu:


src/RMGRunAction.cc
```diff
@@ -99,6 +97,16 @@ void RMGRunAction::BeginOfRunAction(const G4Run*) {
     auto ana_man = G4AnalysisManager::Instance();
     fCurrentOutputFile = BuildOutputFile();
     auto fn = fCurrentOutputFile.first.string();
+
+    // ntuple merging is only supported for some file types. Unfortunately, the function to
+    // check for this capability is private, so we have to replicate this here. Also it can only be
+    // called after opening the file, when setting the flag does not work any more :-(
+    auto file_type = fCurrentOutputFile.first.extension();
+    if (file_type != ".csv" && file_type != ".CSV" && file_type != ".xml" && file_type != ".CSV" &&
```
EricMEsch, line 105: You are checking for `.CSV` twice i think :smiley:

ManuelHu: Ah thanks, should be "XML" 😄

PR #188, opened 2024-12-09
Dockerfile: remove leading v for versions

### LINKED ISSUE #185: Remove leading `v` from version slug in `.dockerhub/Dockerfile`, before passing to CMake

ManuelHu:
fixes #185  ?


gipert:
The issue applies to version slugs fed to the build script as well, when a stable release is tagged. The git tag is passed, and it typically contains a `v` -> we should remove it

ManuelHu:
you mean something like this (we cannot use bash substitution here, that's why I used sed...)

PR #193, opened 2024-12-11
Use double to store energy in hits

ManuelHu:
So we do not pretend that we have more precision in output than we actually have...


gipert:
Uhm this needs some more discussion, we have been using floats in the past to reduce disk space usage and I'm not sure we need double precision...

gipert:
We need doubles only if we need to store very small or very big numbers precisely, so this is not the case for positions (and also energy...?)

gipert:
Sorry I misunderstood, then no problem!

PR #198, opened 2024-12-16
include _t<n> suffix in check for existing files 

### LINKED ISSUE #197: Multithreading for .hdf5 file output
It seems that commit 634ef19ee21b141e155cdaf961c9d074c125d3ee broke multithreading for .hdf5 files. 

Even if there are no files in the output folder, trying to run an application with more than one thread and without the `-w` flag to overwrite output files will crash, because multiple threads will try to create an output file with the same name. It seems like the `_t` extra for each thread created by Geant4 is not included in checking if the file already exists.

It will still work as usual if you specify `-w`, but it seems kinda odd to have the feature if it does not work without it.

ManuelHu:
fixes #197

for root files with ntuple merging, there might still be false-positives. But we do not care that much about root files at the moment :-)

### DIFF
--- a/src/RMGRunAction.cc
+++ b/src/RMGRunAction.cc
@@ -242,7 +242,8 @@ std::pair<fs::path, fs::path> RMGRunAction::BuildOutputFile() const {
   auto path_for_overwrite =
       fs::path(G4Analysis::GetTnFileName(path.string(), path.extension().string()));
   if (fs::exists(path_for_overwrite) && !manager->GetOutputOverwriteFiles()) {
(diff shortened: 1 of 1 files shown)


tdixon97:
Should we just stop supporting the ROOT file output option, just another thing to maintain..

ManuelHu:
We do not really "do" anything to support root. It would take more code for us to disallow root files...

ManuelHu:
This is remarkably complex, and needs more investigation... (apparently, there is no way to abort the run from the worker. And at the master BeginOfRunAction, the application cannot not know the number of threads...)

PR #201, opened 2024-12-18
Setup dummy X server for CI tests

gipert:
- tests/docs dependencies are already available in the base image (see recent changes to https://github.com/gipert/remage-docker)
- use Xvfb to start a dummy X server -> enables Geant4 visualization
- upload test output files matching pattern `*.output.*` to GitHub as artifacts

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,11 +23,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # guideline: only test on latest patch releases (and maybe previous one)
         container_version:
         - G4.11.3
         - G4.11.2
-        - G4.11.1
-        - G4.11.0
         - slim
 
     container: docker://gipert/remage-base:${{ matrix.container_version }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,10 +33,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - name: Install pip dependencies for tests
-      # TODO: replace with a better way to manage test dependencies.
-      run: |
-        pip3 install --user --upgrade legend-pydataobj scipy
+
     - name: Build project
       run: |
         mkdir build
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ concurrency:
 jobs:
 
   test_on_linux:
-    name: Test on Linux containers (Ubuntu LTS)
+    name: Test on remage image
     runs-on: ubuntu-latest
 
     strategy:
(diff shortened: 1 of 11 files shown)


ManuelHu:
I wouldn't remove the test skipping. This should be fine, and not skip the tests when X11 is available, so not strictly required here...?

And running the tests on clusters or servers without X access available (or necessary) should still be easy and supported, in my opinion ("easy" as in not having to remember the ctest options... btw, ctest does not check what is passed as commandline arguments, it happily ignores all invalid/mistyped flags 🙄.)

gipert:
> I wouldn't remove the test skipping. This should be fine, and not skip the tests when X11 is available, so not strictly required here...?

yes, I just wanted to be extra sure for testing. We are leaving it there

Something weird is going on on some containers:

```
 -------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : opengl1001
      issued by : G4OpenGLStoredSceneHandler::AddPrimitivePreambleInternal
Error attempting to create an OpenGL display list.
Current display list id: 257
Maybe out of memory?
*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------
```
and 
```
 GL2PS info: OpenGL feedback buffer overflow
GL2PS info: OpenGL feedback buffer overflow
GL2PS info: OpenGL feedback buffer overflow
GL2PS info: OpenGL feedback buffer overflow
GL2PS info: OpenGL feedback buffer overflow
GL2PS info: OpenGL feedback buffer overflow
```

I will try installing the Mesa stuff

ManuelHu:
The GL2PS warnings can be ignored, they are always there. I never got GL2PS working without warnings, if it works at all...

ManuelHu:
If we could skip the test on G4 11.0, we could use TOOLSSG_OFFSCREEN starting with 11.1:
https://geant4.web.cern.ch/documentation/dev/bfad_html/ForApplicationDevelopers/Visualization/visdrivers.html#visdrv-toolssg

TOOLSSG_OFFSCREEN should also work in slim, where no X11 or OpenGL support is compiled into the base image...

ManuelHu:
11.0 seems also to be broken with a navigaton bug, unrelated to visualization

ManuelHu:
or maybe OGLIX works? this apparently fixed a similar problem once: https://geant4-forum.web.cern.ch/t/opengl-issues-after-recent-ubuntu-update/10967

ManuelHu:
> Warning: No files were found with the provided path: build/tests/**/\*.output.*. No artifacts will be uploaded.

gipert:
Guess what... Geant4 does not want anyone to tell it what it has to do...

```
complex-volume.output_0000.pdf
```

gipert:
Seems to work. Test outputs can be downloaded from here: https://github.com/legend-exp/remage/actions/runs/12416324939

PR #202, opened 2024-12-19
Update developer's guide

gipert:
Also: moved developer guide in a place that GitHub can detect and linked back to docs


ManuelHu:
Hmm, that would be nice if it worked well.

Neither internal links nor the admonitions are rendering correctly [in the MD view on github](https://github.com/legend-exp/remage/blob/4c4401c7ac8ee4507bf3b781de926ad6f1a02639/.github/CONTRIBUTING.md). The section on dependencies is also empty...
Unfortunately there is also no way to use github-style admoitions in sphinx (I already looked into that once)

As this guide would be linked prominently by github, I do not really see the value in this. Having only half of the guide on github would be worse than just having a stub CONTRIBUTING.md with a link to RTD, IMO...

gipert:
I was too optimistic. Let me add a stub CONTRIBUTING.md

ManuelHu (approved):

PR #206, opened 2024-12-20
only store G4TouchableHandle if visualization is enabled

### LINKED ISSUE #205: Memory consumption due to G4TouchableHandle in RMGOpticalDetectorHit
I have tried to rerun my full cosmogenic simulations for the water tank (e.g. Simulate 1e7 muons and output all detector responses for every Ge77 event, including PMTs) and i suddenly started having memory issues. I have not exactly pinned it down, but i am 99% sure it has to be due to commit f0d40db, that added a `G4TouchableHandle` field to `RMGOpticalDetectorHit`.

Here are the heaptrack analytics of the run when i go to commit f0d40db (I ran out of memory before finishing the run): 
After

And here is it if i go back to 172f1a1: 
Before

As the `G4TouchableHandle` there is only required for the `Draw()` function, which has no use in batch mode, is it possible to hide this in batch mode? Or just use the pointer instead of the actual object?

ManuelHu:
fixes #205


ManuelHu:
@EricMEsch could you test if this PR fixes the memory leak you observed?

EricMEsch:
It does not seem to work. I tested and the memory consumption is still the same. I added a print statement inside of `if(G4VVisManager::GetConcreteInstance()` and it seems that if statement is true, even if i run in batch mode. If I run e.g. example 03-optics i will still see the print.

EricMEsch:
This is probably because in `RMGManager::Initialize()` the `fG4VisManager` is Initialized in any case. It might be a solution to just add a variable that will be set upon entering `RMGManager::StartInteractiveSession()`?

ManuelHu:
oh no, but `fInteractive` does not really help here. Visualization can also be used in batch files... (i.e. running a simulation and exporting an image file, like in our CI tests here). I will try some things to get the actual vis state from the manager...

EricMEsch (approved):

PR #207, opened 2024-12-20
cmake: get version from git, or git archive

ManuelHu:
this needs still quite some testing

also it still does not update the version automatically after you add a tag locally...


gipert:
Does #211 help?

ManuelHu:
@gipert should be ready now. Testing instructions:

* download the tarball for this branch from github and try to compile from it, ensure it has the right version output.
* compile from git checkout, ensure it has the right version output (should work, as CI already does this)

note: the version is still marked as `0.6.2-11-g...`, as the release 0.6.3 was done from a different branch - so `git describe` does not take it into account

gipert:
We should make sure that this version always matches with the python package version. At the moment they are determined independently but exactly in the same way? Should we instead determine the version with one of the two methods and then forward it to the other?

Otherwise this looks good to me. Having two versions is ugly but I understand that CMake is picky.

gipert:
For example, it would be great to make CMake run this:

https://setuptools-scm.readthedocs.io/en/stable/usage/#as-cli-tool

to determine the version. The problem though is that setuptools-scm must be installed...

ManuelHu:
> Having two versions is ugly but I understand that CMake is picky.

yes, cmake does unfortunately not accept arbitraty version strings, only `major.minor.patch` (+ _some_ allowed suffixes)

> Should we instead determine the version with one of the two methods and then forward it to the other?

that would be great; but how?

* We could set `SETUPTOOLS_SCM_PRETEND_VERSION` when calling the install command? But that will not work if anyone tries to install the python package by themselves (i.e. outside the default venv created with cmake)
* We could create a `.git-archival.txt` file in the source directory ourselves, and try to make setuptools-scm use that?
* We could also do something similar by ourselves (but this is ugly)

> The problem though is that setuptools-scm must be installed...

And even though the build process does _use_ setuptools-scm, it will just be installed in a temporary buildroot. So it cannot be used outside

cmake/RMGGitVersion.cmake
```diff
@@ -0,0 +1,37 @@
...
+  endif()
+endwhile()
+
+# try to get the tag ourselves from git describe, if we did not have it set above
+# (i.e. we are not in a git-created archive).
+if(${RMG_GIT_VERSION} STREQUAL "v0.0.0" AND IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/.git")
+  find_package(Git)
+  if(Git_FOUND)
+    execute_process(
+      COMMAND ${GIT_EXECUTABLE} describe --tags --match "*[0-9]*"
```
gipert, line 19:
```suggestion
      COMMAND ${GIT_EXECUTABLE} describe --tags --always --match "*[0-9]*"
```

ManuelHu:
hmm, `--always` does not _really_ work. It outputs a commit hash, but that cannot be used by cmake as a version number (that undocumented(?) expected version format...)

ManuelHu:
sneak peak:

```
(python_venv) test /work/remage/build $  remage
  _ __ ___ _ __ ___   __ _  __ _  ___ 
 | '__/ _ \ '_ ` _ \ / _` |/ _` |/ _ \
 | | |  __/ | | | | | (_| | (_| |  __/ v0.8.1.dev1+g0e8ac7d.d20250210
 |_|  \___|_| |_| |_|\__,_|\__, |\___| using geant4-11-02-patch-02 [MT]
                           |___/      

[Summary -> Realm set to DoubleBetaDecay
[Warning -> Could not find G4Region 'SensitiveRegion' in the store. No special production cuts applied
[Summary -> CLHEP::HepRandom seed set to: 457463257
[Fatal ---> Batch mode has been requested but no macro file has been set
terminate called after throwing an instance of 'std::runtime_error'
  what():  A fatal exception has occurred, the execution cannot continue.
(python_venv) test /work/remage/build $  pip list | grep remage
remage                 0.8.1.dev1+g0e8ac7d.d20250210
(python_venv) test /work/remage/build $ 
```

I got setuptools-scm working (hopefully reliably)

tdixon97:
stop asking questions

gipert:
mind your business, this is not your pull request

gipert:
amazing

PR #210, opened 2024-12-28
Infrastructure for Python wrapper

gipert:
- [x] add pyproject.toml and more pre-commit hooks
- [x] first version of infrastructure for Python wrapper
- [x] improve remage python CLI
- [x] more tests
- [x] update documentation
- [x] make CMake detect changes in python stuff


gipert:
I have tagged a new release and will merge this now since we need to move on. We need to test it extensively to make sure it's robust, before releasing it

Issue #56, opened 2023-07-06
Implement sampling vertices from subtractions


gipert:
Useful to simulate e.g. K42 decays outside mini-shroud but inside cylinder.

gipert:
I'm not sure what I meant here, but this seems to me already supported.

ManuelHu:
I suppose you meant "from a substraction of two user-defined solids" (i.e. where none of them is from the original geometry). This is not yet possible

gipert:
Ah yes you are right...

tdixon97:
do we really need this? whats the application? E..g LAr inside MS can be simulated already

gipert:
LAr outside MS?

tdixon97:
Yes I get it now what you want is to select points where A & !B compared to what you have now which is A& B. Although thats slightly different to what Manuel suggested above.
Do we really want to add more features to the RMG generators though, or move to python entirely? For me its seems much easier, and better to do it sooner.

gipert:
In C++ we have `::IsInside()`, that's quite crucial...

tdixon97:
Ok now I am remembering this discussion. I will work on this implementation and meanwhile think of some high level tests we can do.

PR #220, opened 2025-01-03
Test on containment for HPGes's

tdixon97:
Based on generating primaries in each Ge detector for a fake simplified Ge detector array and checking the fraction in each hpge.
Produces a plot of the fraction in each detector and computes p-value (assuming Wilkes thm), test fails if Z > sigma.
To work needs:
- a new realease of pygeom-tools
- I needed to set BUILD_TEST=1 before make so maybe the build needs to be modified.


gipert:
Somehow pyg4ometry is not installed... I'll have a look

tdixon97:
Yes I found that locally too, I had to set the enviroment variable BUILD_TEST=1 before the build.

tests/confinement/test_lar_intersection.py
```diff
@@ -0,0 +1,223 @@
...
+        np.arange(5), [f"string {i}" for i in range(5)], rotation=90, fontsize=13
+    )
+    ax[1].axhline(y=0, color="red")
+    ax[1].grid()
+    fig.suptitle(f"Intersection check ({sigma:.1f} $\sigma$)")
+    plt.tight_layout()
+    pdf.savefig()
+
+
+if sigma < 5:
```
gipert, line 220: have you tried `assert`?

tdixon97: didnt try i thought you needed the sys.exit() call

ManuelHu: unhandled exceptions produce an exit code of 1; and `assert` statements produce exceptions if they fail... so this should work with assertions

tdixon97:
This should now be ready and working for subtractions (this would close #56). 
The sampling mode:

```
/RMG/Generator/Confinement/SamplingMode SubtractGeometrical
``` 

should be used.
Then a  geometrical volume to exclude vertices from is specified with the same syntax as a Geometrical volume except `AddSolid` is replaced by `AddExcludedSolid`.
You can find some examples in tests. 

Some things to keep in mind  / possible improvements (@ManuelHu / @gipert  / @EricMEsch ):
- The handling of intersections and subtractions of surface vertex is not checked and not fully general (I guess we want the possibly to intersect one surface with one volume).
- The syntax to specify Geometrical volumes is very verbose (see example macro) and involves hardcoded numbers. This could maybe be improved using some constants with the GDML file or some basic expression parsing.
- There is not really need for the 3 sampling modes, the block of code for `SubtractGeometrical` can also handle the other cases.
- We should add proper documentation...

tdixon97:
The issue with sampling for surfaces should be now partially fixed:
- Unlike for volumes the order of sampling matters since the first sampling is on the surface while the intersection is based on the bulk. Now the user has to explicitly chose which they want.
- We should add some tests, but I am not sure what's a reasonable application of this functionality until a generic surface sampler is implemented.
- We should consider the option for intersections of two groups of geometric volumes. Making this very generic will get complicated.

PR #222, opened 2025-01-08
Add Particle filter stacking action option

### LINKED ISSUE #218: Add stacking action options
It might be nice to give the user the possibility to kill certain tracks, for example betas or alphas for sources where we only care about gammas. This could be done with a custom stacking action.
An alternative approach would be custom production cuts for different regions/ volumes (sounds more tricky to me).

EricMEsch:
This would fix #218.

I just quickly took `RMGIsotopeFilter` and adjusted it for this. Its neither compile tested or anything (didnt even update the CMakeLists), i just wanted to see if something like this is what you meant @tdixon97?

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -62,9 +62,13 @@ $ cmake -DCMAKE_INSTALL_PREFIX=<optional prefix> ..
 $ make install
 ```
 
-```{tip}
+:::{warning}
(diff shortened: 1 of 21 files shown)


tdixon97:
@EricMEsch I think this is useful and we should try to test and integrate it, maybe you can reopen?

EricMEsch:
Sure, i will have a look tomorrow

Issue #223, opened 2025-01-09
Tests for DistanceToSurface


EricMEsch:
I just read through the tests toby added in #220 to see how i could add a test checking the Geant4 distance output. I realized that I could just use the output tobys tests already produce (namely the `test-confine.lh5` file created by the `confinement/gen-output` test) to read in and check if the Geant4 produced distance equals the one predicted by legendhpges. 

I am unsure now however, if i should just add this test to those tests (as they are under the folder confinement, and this is not really confinement). But on the other hand it seems unnecessary to recreate the same output. And i am not sure how i can add the fixture dependency if they are not part of the same CMakeLists (as in if i would put them in a new folder)?

tdixon97:
This sounds a really good idea! I am not sure what @gipert thinks but I think its better to make a different folder for this tests, since already there is quite a lot in confinement, even if it means reproducing some outputs. 
For checking the distance you don't really need the so many HPGes as I used. So you could make a simpler GDML with just a single detector.

gipert:
I think it's ok to re-run `confinement/gen-output`, since it's rather inexpensive. But in general I agree that we should use fixtures, especially if simulations are costly. Having things defined in multiple CMakeLists.txt should not be a problem, what matters is that they are included in the right orded.

PR #224, opened 2025-01-10
Add distance to surface tests

EricMEsch:
Generates a dummy gdml volume consisting of 4 detectors (one of each type) and simulates 1 eV electrons sampled within those volumes. Then checks if the distance output from Geant4 agrees with the calculation from `legendhpges`. Currently the tolerance is implemented as default 1 nm. Additionally produces a plot that shows the hit profile for each detector.

Regarding the runtime: the 3 tests added here take combined ~ 10 secs to run on my machine, with all of the now 48 tests (excluding the vis tests) taking ~ 80 secs

Related #223 

Edit: Maybe someone have a look over my python code. I copied a lot of stuff from Tobys implementation, but i also changed a lot. I am not that familiar with python so i might have done something in an overcomplicated way

### DIFF
--- a/tests/distances/make_ge_gdml.py
+++ b/tests/distances/make_ge_gdml.py
@@ -1,0 +1,87 @@
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import numpy as np
+import pyg4ometry as pg4
+import pygeomtools as pytools
+from legendhpges import make_hpge
+
+# read the configs
+out_gdml = "gdml/ge-array.gdml"
+det_macro = "macros/detectors-fake.mac"
+config_dict = {}
+for det_type in ["B", "P", "V", "C"]:
+    with Path.open(Path(f"macros/{det_type}99000A.json")) as file:
+        config_dict[det_type] = json.load(file)
+
+
+def add_hpge(lar, reg, angle, radius, z, idx, dtype):
+    x = radius * np.sin(np.deg2rad(angle))
+    y = radius * np.cos(np.deg2rad(angle))
+    logical_detector = make_hpge(config_dict[dtype], name=f"{dtype}{idx}", registry=reg)
+    logical_detector.pygeom_color_rgba = (0, 1, 1, 0.2)
+    physical_detector = pg4.geant4.PhysicalVolume(
+        [0, 0, 0], [x, y, z], logical_detector, f"{dtype}{idx}", lar, reg
+    )
+
+    physical_detector.pygeom_active_dector = pytools.RemageDetectorInfo(
+        "germanium",
+        idx,
+        config_dict[dtype],
+    )
+    return idx + 1
+
+
+# construct geometry
+reg = pg4.geant4.Registry()
+ws = pg4.geant4.solid.Box("ws", 500, 500, 500, reg, lunit="mm")
+wl = pg4.geant4.LogicalVolume(ws, "G4_Galactic", "wl", reg)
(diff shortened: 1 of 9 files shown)


tests/distances/make_ge_gdml.py
```diff
@@ -0,0 +1,87 @@
...
+reg = pg4.geant4.Registry()
+ws = pg4.geant4.solid.Box("ws", 500, 500, 500, reg, lunit="mm")
+wl = pg4.geant4.LogicalVolume(ws, "G4_Galactic", "wl", reg)
+wl.pygeom_color_rgba = (0.1, 1, 0.1, 0.5)
+
+reg.setWorld(wl)
+
+
+# lar
+lar_s = pg4.geant4.solid.Tubs(
```
tdixon97, line 47: lar can be removed from this GDML, but then change the mother volume for HPGe to be the world not lar.

tests/distances/make_ge_gdml.py
```diff
@@ -0,0 +1,87 @@
...
+pg4.geant4.PhysicalVolume([0, 0, 0], [0, 0, 0], lar_l, "LAr", wl, registry=reg)
+
+
+# hpge strings
+string_radius = 85
+string_angles = [0, 90, 180, 270]
+detectors = ["V", "P", "B", "C"]
+
+
+n = 0
```
tdixon97, line 61: these lines are not needed in this case. They only create a macro to simulate in the LAr cylinders.

EricMEsch:
The test failed because the pre-commit hooks (namely ruff apparently) removed the `from legendhpges import draw` import due to "unused import" :facepalm:. I just changed the line to directly use that function now (although importing the submodule would probably work as well).

tests/distances/make_ge_gdml.py
EricMEsch: I do kinda need both lines. The list of detectors is so i can just loop through all 4 types (although i can omit the detectors variable by just directly inserting the list into the enumerator).

And i need the n to keep track of the unique detector IDs to also create the `detectors-fake.mac` to register the detectors.

EricMEsch: Ah yes, you are right. I dont know how i deleted the write statements below but not that^^

EricMEsch:
I just did some tests:

> I think the threshold could be a bit smaller but also 1nm sounds fine.

I could reduce it. The distance in the output file is saved as a double in meters. For double we expect an accuracy of approximately 16 decimals, so we could even go to fm accuracy (although i am not sure of the accuracy in `legendhpges`, but my tests confirmed that they agree on 1e-15 meters tolerance).

The offset of possibly 5 meters should not be relevant. As we can only query the distance function when we are inside the detector and the distance gets calculated in the local frame of reference. So even if we made a geometry where we put a detector at 1 AU distance to the origin, the local frame of reference calculation should be fine (although other parts would make problems at those distances). In the local frame of reference the distance is limited by the size of the detectors.

tdixon97:
Yeah I'm worried about the global offset more for the distances from `legendhpges`, which is based on the global positions. These distances are also validated by this test.

EricMEsch:
Ah, i just realized that of course in geant4 we do encounter the same issue. As we do as well use the global coordinates first to then transform them into the local coordinates. And if we lost accuracy, we can not get it magically back using the transform. So forget my previous example^^

In Geant4 doubles are used, so the accuracy would roughly be `(largest offset) * 1e-16` which should be fine for our setup

gipert:
Awesome, thanks Eric.

PR #226, opened 2025-01-13
Generic surface sampler

tdixon97:
This implements the generic surface sampling following the algorithm of @jasondet et. al's paper [[link]](https://arxiv.org/abs/0802.2960). 
No geantino are used instead only the methods `G4VSolid::DistanceToIn(G4ThreeVector &p,G4ThreeVector &v)` and `DistanceToOut` are needed. These essentially track the distance to the next surface along a line and so can give us the number of intersections.
I started on testing.
- [x] unit tests (of the C++ code) for the basic methods (this could be expanded to more types of volume),
- [x] some visualisation (it was tricky to get this to work well), for different simple shapes,
- [x] also some plots in python of the vertices and a check of selecting points on each surface and checking the fraction is equal to the ratio of surface areas,
The last test is implemented but for some reason fails for the Union of 2 G4Box, in addition I find some generated points not on the surface, I will investigate.
**Update** : This was related to the origin of the bounding sphere which wasn't accounted for - is now fixed.

Once we have this sorted we should think how to speed it up for difficult to sample solids, one idea is to check whether its impossible for a direction vector to intersect with the solid (eg. it points the wrong way) and then avoid tracking.


gipert:
Should we start adding some docstrings to the important functions? Can you give it a try? https://www.doxygen.nl/manual/docblocks.html

Please also add a link to the paper.

src/RMGVertexConfinement.cc
```diff
@@ -121,8 +121,125 @@ bool RMGVertexConfinement::SampleableObject::IsInside(const G4ThreeVector& verte
...
+std::vector<G4ThreeVector> RMGVertexConfinement::SampleableObject::GetIntersections(G4ThreeVector start,
+    G4ThreeVector dir) const {
+
+
+  // check if the physical volume exists
+  if ((not this->physical_volume) or (not this->physical_volume->GetLogicalVolume()->GetSolid()))
+    RMGLog::OutDev(RMGLog::fatal, "Cannot find number of intersections for a SampleableObject where the physical volume is not set",
+        "this probably means you are trying to generically sample a geometrical volume which "
+        "instead",
+        "should be natively sampled");
```
gipert, line 136: amazing linebreaks

src/RMGVertexConfinement.cc
```diff
@@ -121,8 +121,125 @@ bool RMGVertexConfinement::SampleableObject::IsInside(const G4ThreeVector& verte
...
+    dist = (counter % 2) == 0 ? solid->DistanceToIn(int_point, dir)
+                              : solid->DistanceToOut(int_point, dir);
+
+    int_point = int_point + dist * dir;
+
+    if (dist < kInfinity) intersections.push_back(int_point);
+    counter++;
+  }
+
+  if (intersections.size() % 2) throw;
```
gipert, line 161: Just `throw`???

src/RMGVertexConfinement.cc
```diff
@@ -121,8 +121,125 @@ bool RMGVertexConfinement::SampleableObject::IsInside(const G4ThreeVector& verte
...
+      physical_volume->GetLogicalVolume()->GetSolid()->GetExtent().GetExtentRadius();
+
+
+  // start on z-axis, pointing down.
+  pos = G4ThreeVector(0.0, 0.0, bounding_radius);
+  dir = G4ThreeVector(0.0, 0.0, -1.0);
+
+  // push in rho direction by some impact parameter
+  G4double diskPhi = 2.0 * CLHEP::pi * G4UniformRand();
+  G4double diskR = sqrt(G4UniformRand()) * bounding_radius;
```
gipert, line 189:
```suggestion
  G4double disk_phi = 2.0 * CLHEP::pi * G4UniformRand();
  G4double disk_r = sqrt(G4UniformRand()) * bounding_radius;
```
we should use snake casing, except for class members where we use `fMember`.

src/RMGVertexConfinement.cc
```diff
@@ -121,8 +121,125 @@ bool RMGVertexConfinement::SampleableObject::IsInside(const G4ThreeVector& verte
...
+
+  pos += G4ThreeVector(cos(diskPhi) * diskR, sin(diskPhi) * diskR, 0);
```
gipert, line 191: this can be done with some pre-defined transformation function? https://cmd.inp.nsk.su/old/cmd2/manuals/cernlib/CLHEP/RefGuide/Vector/Hep3Vector.html

src/RMGVertexConfinement.cc
```diff
@@ -121,8 +121,125 @@ bool RMGVertexConfinement::SampleableObject::IsInside(const G4ThreeVector& verte
...
+
+  // now rotate pos and dir by some random direction
+  G4double theta = acos(2.0 * G4UniformRand() - 1.0);
+  G4double phi = 2.0 * CLHEP::pi * G4UniformRand();
+
+  pos.rotateY(theta);
```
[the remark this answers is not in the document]
tdixon97: I took this from Mage we should check if it works well and also test it

src/RMGVertexConfinement.cc
tdixon97: Oops that was left in from when I was testing this outside of remage

src/RMGVertexConfinement.cc
tdixon97: blame clang-format

tests/confinement/CMakeLists.txt
```diff
@@ -94,3 +93,98 @@ add_test(NAME confinment-lar/intersection-and-subtraction
 
 set_tests_properties(confinment-lar/intersection-and-subtraction
                      PROPERTIES LABELS extra FIXTURES_REQUIRED confine-lar-ibn-out-output-fixture)
+
+# test generic surface sampling methods
+find_package(BxDecay0 ${RMG_BXDECAY0_MINIMUM_VERSION} REQUIRED COMPONENTS Geant4)
+
+add_executable(test-surface-sampler-methods EXCLUDE_FROM_ALL test-surface-sampler-methods.cpp)
+target_link_libraries(test-surface-sampler-methods PUBLIC remage BxDecay0::BxDecay0)
```
ManuelHu, line 101: why are you linking against bxdecay0? That should not be neccessary here (as `libremage` already links against bxdecay0, if applicable?

This is also why the tests in the slim container are still failing (that has no bxdecay0 available)

tdixon97: I needed to link with BxDecay0 or RMGVertexConfinement wouldn't compile

ManuelHu: ah yes, that makes sense (as the `RMGVVertexGenerator` uses a bxdecay0 header)

then please link it conditionally, as everywhere else in remage :-)

tdixon97:
> Should we start adding some docstrings to the important functions? Can you give it a try? https://www.doxygen.nl/manual/docblocks.html
> 
> Please also add a link to the paper.

I agree but there seems to be many options, maybe lets pick one ?

tdixon97:
The number of PDF produced by the tests is now large. I think we should consider how to organise better the various confinement tests but also perhaps a strategy to merge the figures into one document?

jasondet:
Yes, point in the +Z direction, then rotate about y by theta then about Z
by phi to get a vector in the direction theta phi.

-------------------------------------
Jason Detwiler
Department of Physics, Box 351560
University of Washington
Seattle, WA 98195
(206) 543-4054

gipert:
>The number of PDF produced by the tests is now large. I think we should consider how to organise better the various confinement tests but also perhaps a strategy to merge the figures into one document?

Yes, please feel free to reorganize things. If you come up with a proposal on how to merge things, I can implement something quick

gipert:
>I agree but there seems to be many options, maybe lets pick one ?

I like the first one:

```
/**
 *
 */
```

tests/confinement/CMakeLists.txt
```diff
@@ -94,3 +93,102 @@ add_test(NAME confinment-lar/intersection-and-subtraction
...
```
ManuelHu, line 101:
```suggestion
target_link_libraries(test-surface-sampler-methods PUBLIC remage)
```

you still need to remove that here

src/RMGVertexConfinement.cc
```diff
@@ -121,8 +121,129 @@ bool RMGVertexConfinement::SampleableObject::IsInside(const G4ThreeVector& verte
...
+  std::vector<G4ThreeVector> intersections = {};
+  G4ThreeVector int_point;
+
+  int_point = start;
+  int counter = 0;
+
+  dist = 0;
+
+  while (dist < kInfinity) {
+
```
gipert, line 150: all these double newlines are not needed here
```suggestion

// get the intersections between the solid and an arbitrary line, defined by start and dir
std::vector<G4ThreeVector> RMGVertexConfinement::SampleableObject::GetIntersections(G4ThreeVector start,
    G4ThreeVector dir) const {

  // check if the physical volume exists
  if ((not this->physical_volume) or (not this->physical_volume->GetLogicalVolume()->GetSolid()))
    RMGLog::OutDev(RMGLog::fatal, "Cannot find number of intersections for a SampleableObject ",
        "where the physical volume is not set this probably means you are trying to generically ",
        "sample a geometrical volume which instead should be natively sampled");

  auto solid = this->physical_volume->GetLogicalVolume()->GetSolid();

  G4double dist = 0;
  std::vector<G4ThreeVector> intersections = {};
  G4ThreeVector int_point;

  int_point = start;
  int counter = 0;

  dist = 0;

  while (dist < kInfinity) {
```

src/RMGVertexConfinement.cc
```diff
@@ -121,8 +121,129 @@ bool RMGVertexConfinement::SampleableObject::IsInside(const G4ThreeVector& verte
 
   return false;
 }
+
+
+// get the intersections between the solid and an arbitrary line, defined by start and dir
+std::vector<G4ThreeVector> RMGVertexConfinement::SampleableObject::GetIntersections(G4ThreeVector start,
```
gipert, line 127: could you add more comments to the core parts of the algorithm?

tdixon97:
> > The number of PDF produced by the tests is now large. I think we should consider how to organise better the various confinement tests but also perhaps a strategy to merge the figures into one document?
> 
> Yes, please feel free to reorganize things. If you come up with a proposal on how to merge things, I can implement something quick

What about making a LaTex document with all the plots, although then of course we need some compiler in the container, and to constantly update that template.

gipert:
Depends on how you want this report document to look like. If it's just a collection of plots then we can do something more simple than LaTeX...

tdixon97:
Well I think it would also be good to have some headings.

tdixon97:
@jasondet Your algorithm generates N intersections and then picks one (or zero).
I understand that formally this keeps the sampled points independent. But I wonder if its really necessary for our applications.
In particular, this is a bit challenging since the maximum number of intersections has to be known in advance.

tdixon97:
My geant4 visualisations don't work properly in CI, fine locally...

tests/confinement/CMakeLists.txt
```diff
@@ -94,3 +93,102 @@ add_test(NAME confinment-lar/intersection-and-subtraction
...
+    test-intersections-union
+    test-containment-subtraction
+    test-containment-union)
+
+foreach(_test ${_surface_tests})
+
+  add_test(NAME confinment-surface/${_test} COMMAND ./test-surface-sampler-methods ${_test})
+  set_tests_properties(confinment-surface/${_test} PROPERTIES LABELS extra FIXTURES_REQUIRED
+                                                              test-surface-sampler-methods-fixture)
+endforeach()
```
ManuelHu, line 133:
```suggestion
add_executable(test-surface-sampler-methods EXCLUDE_FROM_ALL test-surface-sampler-methods.cpp)
target_link_libraries(test-surface-sampler-methods PUBLIC remage)
if(BxDecay0_FOUND)
  target_link_libraries(test-surface-sampler-methods PRIVATE BxDecay0::BxDecay0_Geant4)
endif()

add_test(NAME confinement/test-surface-sampler-methods-build
         COMMAND "${CMAKE_COMMAND}" --build "${CMAKE_BINARY_DIR}" --config "$<CONFIG>" --target
                 test-surface-sampler-methods)

set_tests_properties(confinement/test-surface-sampler-methods-build
                     PROPERTIES LABELS extra FIXTURES_SETUP test-surface-sampler-methods-fixture)

# add the tests
set(_surface_tests
    test-points-subtraction
    test-points-basic
    test-points-union
    test-intersections-basic
    test-intersections-subtraction
    test-intersections-union
    test-containment-subtraction
    test-containment-union)

foreach(_test ${_surface_tests})

  add_test(NAME confinment-surface/${_test} COMMAND ./test-surface-sampler-methods ${_test})
  set_tests_properties(confinment-surface/${_test} PROPERTIES LABELS extra FIXTURES_REQUIRED
                                                              test-surface-sampler-methods-fixture)
endforeach()
```

it really should be enough to conditionally link BxDecay0. The test does not use it itself, only via the header files (which do not include Bxdecay if it is not present)

also, we always link against `BxDecay0::BxDecay0_Geant4` and not `BxDecay0::BxDecay0`

jasondet:
@tdixon97 it is indeed probably not necessary to pick one of the N crossings at random -- at high stats you still get distribution that is essentially uniform over any surface, the only price you pay is that some consecutive points have some hint of a correlation (they lie along the same ray). That kind of correlation is probably irrelevant for our application (rad sims).

However inputting N_max is not difficult. We would just set it to something that seemed reasonable (like 16 or 32 or something) and then bumped it up by a factor of 2 and re-ran if we ever saw the error message that it was too small. Then we didn't have to worry about arguing whether the faint correlation was okay or not. So we opted to do that.

tdixon97:
> @tdixon97 it is indeed probably not necessary to pick one of the N crossings at random -- at high stats you still get distribution that is essentially uniform over any surface, the only price you pay is that some consecutive points have some hint of a correlation (they lie along the same ray). That kind of correlation is probably irrelevant for our application (rad sims).
> 
> However inputting N_max is not difficult. We would just set it to something that seemed reasonable (like 16 or 32 or something) and then bumped it up by a factor of 2 and re-ran if we ever saw the error message that it was too small. Then we didn't have to worry about arguing whether the faint correlation was okay or not. So we opted to do that.

I wrote our implementation so only one G4VSolid is ever sampled at a time Nmax can probably be usually something around 6 in our case (so less bad). However, since we weight the intersections by N/Nmax this still means we though away a lot of points and so are also using a lot of calls to `GetRandom()`. I guess that the vertex confinement is usually so fast this doesn't matter. But there are cases where it might...
I agree the correlation is clearly there but probably minimal.

Testing/Temporary/LastTest.log
gipert, line 1: remove this

docs/developer.md
```diff
@@ -62,9 +62,13 @@ $ cmake -DCMAKE_INSTALL_PREFIX=<optional prefix> ..
 $ make install
 ```
 
-```{tip}
+:::{warning}
+If you want to run the _remage_ tests the cmake flag `-DBUILD_TESTING=ON` is required.
```
gipert, line 66: but the default should be `ON`... maybe mention it.

include/RMGVertexConfinement.hh
```diff
@@ -91,33 +111,140 @@ class RMGVertexConfinement : public RMGVVertexGenerator {
...
+         *
+         * @param v The physical volume.
+         * @param r A rotation matrix for the sampling solid.
+         * @param t A translation vector for the sampling solid.
+         * @param s A solid for geometrical volume sampling or for generating candidate points for
+         * rejection sampling.
+         * @param ns A flag of whether the solid is natively sampeable.
+         * @param ss A flag of whether the solid should be sampled on the surface.
+         */
         SampleableObject(G4VPhysicalVolume* v, G4RotationMatrix r, G4ThreeVector t, G4VSolid* s,
```
gipert, line 136: should we use better argument names? I don't remember why I chose these stupid ones...

include/RMGVertexConfinement.hh
```diff
@@ -91,33 +111,140 @@ class RMGVertexConfinement : public RMGVVertexGenerator {
...
     }
 
+    /**
+     * An object which we can generate position samples in. Based on either a
+     * @c G4VPhysicalVolume or geometrical volume defined by a @c G4VSolid . The
+     * sampling can be performed either on the surface or in the volume of the solid.
+     */
     struct SampleableObject {
 
+
```
gipert, line 121:
```suggestion
```

include/RMGVertexConfinement.hh
```diff
@@ -91,33 +111,140 @@ class RMGVertexConfinement : public RMGVVertexGenerator {
...
+         *
+         * @param start The starting vector of the line, note this should be outside the solid.
+         * @param dir   The direction vector.
+         *
+         * @returns A vector of the points of intersection.
+         */
+        std::vector<G4ThreeVector> GetIntersections(const G4ThreeVector start,
+            const G4ThreeVector dir) const;
+
+
```
gipert, line 204:
```suggestion

```

include/RMGVertexConfinement.hh
```diff
@@ -179,6 +306,7 @@ class RMGVertexConfinement : public RMGVVertexGenerator {
     bool fOnSurface = false;
     bool fForceContainmentCheck = false;
     bool fLastSolidExcluded = false;
+    int fMaxNumIntersections = -1;
```
gipert, line 309: Like for the macro command, let's use a more precise name. Like adding `GSS` in the name

src/RMGVertexConfinement.cc
```diff
@@ -121,8 +121,132 @@ bool RMGVertexConfinement::SampleableObject::IsInside(const G4ThreeVector& verte
 
   return false;
 }
+
```
gipert, line 124:
```suggestion
```

src/RMGVertexConfinement.cc
```diff
@@ -847,6 +1008,16 @@ void RMGVertexConfinement::DefineCommands() {
       .SetStates(G4State_PreInit, G4State_Idle)
       .SetToBeBroadcasted(true);
 
+
+  fMessengers.back()
+      ->DeclareProperty("MaxNumberOfIntersections", fMaxNumIntersections)
```
gipert, line 1013: Same comment about more precise name

tests/tex/CMakeLists.txt
gipert, line 1: mmmmh this is not how it should be done, it should not be a test... Maybe leave this out and I do i properly in a future PR

tests/tex/validation-report.tex
gipert, line 1: I would also leave this out. Future PR one we have decided how to make this report.

gipert (changes requested):

docs/developer.md
tdixon97: are you sure, I had errors running the tests (pyg4ometry didnt get installed) unless I included this.

tdixon97:
Should now be ready @gipert

PR #229, opened 2025-01-16
Add vis label to all tests opening a GUI

EricMEsch:
Some of the tests added did not have the "vis" label although they opened a GUI. When i run tests localy inside of the apptainer container that use the OGL GUI they brick my screen, which means i typically have to restart the computer, so I typically run `ctest -LE vis`. I am sure there is some way to fix my local issue, but i think also the labels should be applied correctly ;)

### DIFF
--- a/tests/confinement/CMakeLists.txt
+++ b/tests/confinement/CMakeLists.txt
@@ -128,8 +128,16 @@ set(_surface_tests
 foreach(_test ${_surface_tests})
 
   add_test(NAME confinment-surface/${_test} COMMAND ./test-surface-sampler-methods ${_test})
-  set_tests_properties(confinment-surface/${_test} PROPERTIES LABELS extra FIXTURES_REQUIRED
-                                                              test-surface-sampler-methods-fixture)
+
+  if(_test MATCHES "test-points-.*")
+    set_tests_properties(
+      confinment-surface/${_test} PROPERTIES LABELS "extra;vis" FIXTURES_REQUIRED
+                                             test-surface-sampler-methods-fixture)
+  else()
+    set_tests_properties(
+      confinment-surface/${_test} PROPERTIES LABELS extra FIXTURES_REQUIRED
+                                             test-surface-sampler-methods-fixture)
+  endif()
 endforeach()
 
 # now run some simulations and test the outputs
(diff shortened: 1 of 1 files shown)


tdixon97:
@gipert is there no way to prevent a GUI being produced?

EricMEsch:
There certainly is, it depends on what you actually want the tests to do. You could force tcsh similar to how its done in the `RMGManager` `std::make_unique<G4UIExecutive>(fArgc, fArgv, val.empty() ? "tcsh" : "")`. This would still open Geant4 in an interactive mode, but without the GUI. I have to admit, i did not look at what the tests exactly do, just at the output they produce. So i am not exactly sure what you need/want.

ManuelHu:
> std::make_unique<G4UIExecutive>(fArgc, fArgv, val.empty() ? "tcsh" : "").  This would still open Geant4 in an interactive mode, but without the GUI.

I would gues this will not change much, at least for the tests. The test are not interactive; so this would not change anything for them...

> is there no way to prevent a GUI being produced?

probably not, at least no nice one in the current framework and without removing our test output. The tests use Xorg-based visualization, and probably there is no way to turn the window off with that. **You can only fully disable visualization, but then you also do not get the rendered PDF output files.**

Some tips for @EricMEsch :
* Have you tried using `export LIBGL_ALWAYS_SOFTWARE=1` in bash before running the tests? I had some success in the past to circumvent some ugly driver crashes with this (I assume you are using linux...)
* One rather ugly solution is to set up `Xfvb` inside the container and let remage use that (by setting `$DISPLAY` appropriately). See [the github CI workflow on how to do that](https://github.com/legend-exp/remage/blob/main/.github/workflows/main.yml#L46-L51)
  then geant4 would not create windows in your real desktop session, but in this virtual one. But getting this to work well can be rather tricky.

and for us all:
* In  the long term, we might want to migrate our tests to `TOOLSG`-based visualization. That toolkit is rather new and had recently added a new option to do offscreen rendering and exporting (in G4 11.2 or so?). but it does not work on all Geant4 11.x versions.

tdixon97:
Another related issue is that many visulisation plots do not get rendered properly when run in CI, I am not sure why...

gipert:
@ManuelHu these tips could go in the dev guide...

EricMEsch:
@ManuelHu thanks for the tipps, `export LIBGL_ALWAYS_SOFTWARE=1` seems to work for me!
I have so far been using a wierd work-around by ssh-ing into myself with X11 forwarding which also used to work.

PR #236, opened 2025-01-23
add a mutex for G4GeneralParticleSource

ManuelHu:
GPS uses global state to store source defintion. It has internal locks, but those (obviouly) do not prevent races between "setting/changing state" and generating a vertex from that state.

* https://github.com/Geant4/geant4/blob/32390e802bd84a8b85ac02a90a3d31cf205dd5b7/source/event/src/G4GeneralParticleSource.cc#L49-L51
* https://github.com/Geant4/geant4/blob/32390e802bd84a8b85ac02a90a3d31cf205dd5b7/source/event/src/G4GeneralParticleSourceData.cc#L63-L68

We already have instances of `G4GeneralParticleSource` per thread (and need those), but `fParticleSource->GetCurrentSource()` is **shared globally**. So we need to lock around this to ensure that every position we set is also only used in one event (or used at all).

In my tests (*) there was quite a "large" probability to observe this race - something like 0.1% - 1%. But that is obviously for artificially slow "actual" simulation time per event. The tests also covering this will be part of my analysis reader PR
The observable symptom was a **random skipping/duplication of generated vertex positions**.

This bug is as old as remage. :-(

### DIFF
--- a/include/RMGGeneratorGPS.hh
+++ b/include/RMGGeneratorGPS.hh
@@ -18,6 +18,7 @@
 
 #include <memory>
 
+#include "G4AutoLock.hh"
 #include "G4GeneralParticleSource.hh"
 #include "G4ThreeVector.hh"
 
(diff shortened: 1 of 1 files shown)


EricMEsch:
The test that fails is my distance test. I am not so sure about the `DeprecationWarning`, as i dont really understand the issue with e.g. ` zlocal = np.array(1000 * steps.zloc)`. Thats not why it fails, it fails cause the distances do not match anymore. As your PR should not change anything on the output side, i expect that the tolerance is too low? 

From Geant4 side with doubles i would expect an accuracy about `(largest offset) * 1e-16`, which should be fine for the test, as there the largest offset is less than ~ 1m and the tolerance is 1e-12 m. Although we were worried about the accuracy of the `legendhpges` package. 

It is kinda odd though, that a mismatch would be this rare then, that it only happened now. Also it is kinda annoying that the output plots are not produced if the tests fail. My code should produce the plots in any case and only raise the error after saving the plots.

tdixon97:
@ManuelHu did you test this? We should definetly add some tests to the CI

ManuelHu:
> did you test this? We should definetly add some tests to the CI

there will be tests with the new vertex reader. I have implemented your idea in those tests (i.e. using one input column as some sort of "index"). With that we can check that there are no duplicate coordinates in the output.

(that's also how I found the bug in the first place)

ManuelHu:
@EricMEsch bd499ea4d6498947ea5fb9432fa0c55d61f86e57 should fix the CI upload issue. lets see now if your test reproducible fails...

ManuelHu:
>  I am not so sure about the DeprecationWarning, as i dont really understand the issue with e.g.  zlocal = np.array(1000 * steps.zloc).

rather off-topic, but you can use `1000 * steps.zloc.to_numpy()` to remove the warnings. This happens apparently since numpy 2.x (Toby did already find and fix this in reboost)

PR #237, opened 2025-01-23
remove implicit conversion to numpy arrays in ge-distance test

EricMEsch:
Question: Should i also increase the tolerance? As it seems that in rare cases this test might fail. It is not 100% certain that it was due to the tolerance, but i guess that is most likely. Currently 1e-9 mm means 1 picometer accuracy, that we do not need.


EricMEsch:
Pretty sure. In the Pipeline it says 
grafik

are_distances_good is a variable that will be false if one of the calculated distance differences is above tolerance.

ManuelHu:
I think we can leave the tolerance as the failure appears very rarely. And we might see regressions earlier if the test starts to fail more often.

Issue #174, opened 2024-12-03
Review maximum step length values and production cuts


ManuelHu:
It's very important to review these values, which were blindly imported from MaGe, and think about how to check their impact (related: #162). They should be set as high as possible, to reduce computing time and output size.

Relevant source code:

https://github.com/legend-exp/remage/blob/2202460fcc0a35afd4a89f5e95d41c4364d94df4/src/RMGPhysics.cc#L355

https://github.com/legend-exp/remage/blob/2202460fcc0a35afd4a89f5e95d41c4364d94df4/src/RMGPhysics.cc#L397

We should also use different cuts for different volumes, if possible!

gipert:
Related: I have been studying the 0vbb containment efficiency vs dead-layer size. Here is a plot (y = containment efficiency, x = dead-layer thickness in mm):

<img src="https://github.com/user-attachments/assets/21739ba4-625e-407d-8a9d-e2e63beead6e" alt="Image description" width="400" />

I implemented the dead layer as practically removing vertices and steps in it -> i.e. completely equivalent to "shrinking" the detector. The expected plot is: monotonically decreasing efficiency vs thickness. As you can see, in practice, the efficiency increases until ~1 mm, before starting to decrease. This might be related to non-adequate production cuts and we should investigate it.

gipert:
We should also discuss ideas to tune the cut values in specific regions of the detectors, like close to the surface, without introducing new sensitive volumes.

EricMEsch:
How exactly did you cut away steps? As I do not understand how the efficiency can go up by removing energy depositions? 

> I implemented the dead layer as practically removing vertices and steps in it

You are observing the same data, but just change something in the readout?

Because if you "shrinked" the detector and changed something about the physical volume, there might be an unforeseen Issue with production-cuts. But if you kept everything exactly the same, and only filtered out some vertices in the read-out, i do not understand (from a point of causality) how it can go up?

gipert:
This is the full script, for reference (also an interesting example of quick post-processing of the single-table remage output). I have heavily validated so i think there is no bug.
```py
import re
import gc
import glob
from concurrent.futures import ThreadPoolExecutor

import matplotlib.pyplot as plt
import hist
import awkward as ak
import numpy as np
import pyg4ometry

from legendhpges import make_hpge
from lgdo import lh5

plt.rcParams["figure.figsize"] = (12, 4)
plt.rcParams["figure.dpi"] = 96

from legendmeta import LegendMetadata

[...]

Issue #43, opened 2023-05-23
Review physics list


gipert:
- compare current implementation with latest Geant4 recommendations
- check if angular correlations are correctly enabled

EricMEsch:
Isnt this more or less the same as #180? Also your results comparing the cosmogenics to MaGe should be part of this? @MoritzNeuberger

gipert:
Yes, we have never enabled hadronic physics in the past (MaGe) and it's fine. We want to keep it as is now, i.e. opt-in

MoritzNeuberger:
@EricMEsch, with the issue in the modular physics list regarding neutron captures, one should look into what the current G4-suggested standard is for implementing them and keep remage up to date on that.

EricMEsch:
While i do like the current way of just specifying basic physicslists (like shielding) i agree, that it seems to be more reasonable to just register all processes yourself? So that you can keep better track of what is happening.

Like we should pick out a set of processes for default remage (and then we can argue about which options we want to give to the user)

gipert:
yes i agree that here we should be verbose and include/exclude processes in detail, instead of relying on some "bundled" G4 method which could expose us to unexpected changes depending on the G4 version

ManuelHu:
but we also need to take into account that these constructors are the "official" way of constructing physics lists in Geant4, so we would need to follow the upstream changes much more closely.

MoritzNeuberger:
I agree with Manuel, also considering that we can only really compare our results with other experiments/simulation stacks when we use the same models. And the usual way people quote what they did is by stating the default physics list they used.

gipert:
This could be also closed?

PR #238, opened 2025-01-23
clang-tidy: more fixes and enable checks

ManuelHu:


include/RMGPhysics.hh
```diff
@@ -73,7 +73,6 @@ class RMGPhysics : public G4VModularPhysicsList {
         double positron;
         double proton;
         double alpha;
-        double muon;
```
ManuelHu, line 76: @gipert this appears to be unused (neither written nor read). Or should it be implemented?

gipert: This should be implemented. We'll get to this when addressing #43 and #174

ManuelHu: I guess there is a reason that this is not implemented:

https://github.com/Geant4/geant4/blob/master/source/processes/cuts/src/G4ProductionCuts.cc#L147-L155

there just is no production cut for muons...

tdixon97: I guess it makes sense, not many processes make muons as secondaries

PR #239, opened 2025-01-23
Analysis manager refactoring

### LINKED ISSUE #156: Expand vertex file reader to parse particle type and momenta
This would make remage support a wide range of external event generators through `RMGVertexFromFile`. The implementation would need to be a bit more complex than what it is right now, since the input file structure is "jagged". I guess a lot of this is already implemented in `RMGGrabmayrGCReader`.

Actually, "vertex" in `RMGVertexFromFile` was originally intended to refer only to vertex position. We should maybe implement all of this as an additional generator.

ManuelHu:
this got larger then anticipated

* global locking for `RMGAnalysisManager`
* thread-safe interface `RMGAnalysisManager::Access` that automatically handles locking and using the correct ntuple id
* add a **simple** `RMGGeneratorFromFile` to read kinetics data. fixes #156 
* adds more extensive tests for input (see also #236 for a bug discovered with them)

---

implemented columns:

* `g4_pid` is the pdg encoded particle
* `px`/`py`/`pz` are a unit vector
* `ekin` is kinectic energy (in MeV)

storing the `g4_evtid` in the input is not possible: changing event id at runtime is not possible. and the only possibility is reading one row at a time, so we cannot seek back/forth until we find a corresponding entry.

an extension to multiple particles / global start times / etc. should be easy, after the infrastructure is now there; I only implemented the most basic input fields right now.

also the combination of position and kinetics vertices will not be stable between runs, and there is also no nice way. The best option would be to (optionally) also read position data from the `kin` table and use the `kUnconfined` mode.

---

this PR does not yet contain adjustments to the MUSUN reader.  combining the MUSUN reader with any other reading task might (quite certainly _will_) fail in weird ways - but this is not possible from macros, at least.


EricMEsch:
> this PR does not yet contain adjustments to the MUSUN reader. combining the MUSUN reader with any other reading task might (quite certainly will) fail in weird ways - but this is not possible from macros, at least.

I think the best option is to leave the MUSUN reader untouched anyways?

tdixon97:
I think so, for now, in the long run I think the best solution is rewritting these generators as external python packages (also MAURINA) making it much easier to test

include/RMGAnalysisReader.hh
```diff
@@ -18,15 +18,86 @@
...
+#include "G4AutoLock.hh"
 #include "G4ThreeVector.hh"
 #include "G4VAnalysisReader.hh"
 
 #include "RMGVVertexGenerator.hh"
 
-class RMGAnalysisReader {
+/**
+ * @brief wrapper around \ref(G4VAnalasisReader) instances with special handling for LH5 files.
+ *
```
EricMEsch, line 29: > \ref(G4VAnalasisReader)

Should probably be `G4VAnalysisReader?`

include/RMGAnalysisReader.hh
```diff
@@ -18,15 +18,86 @@
...
+ * @details notes for threadsafe use:
+ * - opening/closing can only be performed on the master thread.
+ * - in a multithreaded application, all function calls are guarded by a mutex. Worker
+ * threads can use the reader instance after opening, but only one worker can use the reader at a
+ * time.
+ * - the reader access handles are generally thread-safe, if no other thread uses any
+ * \ref(G4VAnalasisReader) directly.
```
EricMEsch, line 36: Same here

ManuelHu: thanks!

src/RMGGeneratorFromFile.cc
```diff
@@ -0,0 +1,122 @@
...
+namespace u = CLHEP;
+
+RMGAnalysisReader* RMGGeneratorFromFile::fReader = new RMGAnalysisReader();
+
+RMGGeneratorFromFile::RMGGeneratorFromFile() : RMGVGenerator("FromFile") {
+  this->DefineCommands();
+  fGun = std::make_unique<G4ParticleGun>();
+}
+
+void RMGGeneratorFromFile::OpenFile(std::string& name) {
```
tdixon97, line 23: is there a way to also use the units in the LH5

ManuelHu:
> I think the best option is to leave the MUSUN reader untouched anyways?

If we have anything else reading a csv file via the `RMGAnalysisReader`, the musun reader would break in the current state. This is because the `G4*AnalysisReader::Instance()` calls do not actually return new instances, but shared global ones. So ntuple numbers might conflict between the readers, for example. And a lock might needed to avoid conflicts.

But: The only two users of the `RMGAnalysisReader` are _alternatives_ to the MUSUN generator, so there is no useful way where you would get this combination right now.

tests/vertex/macros/pos-lh5.mac
```diff
@@ -2,12 +2,12 @@
...
+/RMG/Generator/Confinement/FromFile/FileName macros/vtx-pos.lh5
 
 /RMG/Generator/Select GPS
 /gps/particle     ion
 /gps/ion          81 208
 /gps/energy       0 keV
 /gps/ang/type     iso
 
-/run/beamOn 10
+/run/beamOn 1000
```
tdixon97, line 13: is 1000 enough?

EricMEsch:
> In the long run I think the best solution is rewritting these generators (also MAURINA) as external python packages, and then using the functionality Manuel has created

Just to make sure that i correctly understand this: this means that the functionality that will be added here allows to read in primary vertices AND GammaCascades with this AnalysisReader (at least from .lh5 and .hdf5) at the same time during runtime?

src/RMGGeneratorFromFile.cc
ManuelHu: Hmm, no simple way at least. except if we want to implement a full unit system in C++. (maybe we could re-use some GDML expressions for this 😅🫣)

the current "stupid" way is to append the unit to the column name (like in our metadata). but that is not useful, because there is no way to test for the availability of columns....

ManuelHu:
this reader should allow to read any number of files at runtime (formats: csv, xml, hdf5, lh5, root) - at the cost of global locking across all file accesses.

but it would not be a drop-in replacement for your hand-coded gamma cascade reader. the analysis reader can only read row-by-row. no going back, no arbitrary seeking, not even a way to reset to the first row; and also no access to meta information (i.e. number of columns, available rows in the file, ...)

src/RMGGeneratorFromFile.cc
ManuelHu: but I can at least prepare something more helpful

tdixon97: At minimum we should then document clearly what is the expected unit.

EricMEsch:
Okay, that sounds great. Sure, this is not a drop-in replacement, I was just wondering if this technically would allow to replace the hand-coded reader later down the line. The only thing one has to think about then is how to structure the gammacascades to deal with these restrictions (which might get pretty complicated and annoying).

ManuelHu:
as the number is low, you could probably read all lines into an array _once_ (i.e. before the run starts). Then you can index into that array at runtime.

src/RMGGeneratorFromFile.cc
ManuelHu: I am now preparing a _very_ simple (and very manual) form of unit handling.

tdixon97:
> Okay, that sounds great. Sure, this is not a drop-in replacement, I was just wondering if this technically would allow to replace the hand-coded reader later down the line. The only thing one has to think about then is how to structure the gammacascades to deal with these restrictions (which might get pretty complicated and annoying).

I think the best solution is to move all that code to python and create an array of kinematics to read into remage. Our philosophy should be to do things in python or julia whenever possible. But this is for the future

EricMEsch:
I will test the musun reader as soon as i find time

tdixon97:
For me it looks ok, but will need testing. Once its merged I can try running some simulations.

EricMEsch:
The musun generator does not seem to be able to read the musun files. I first figured out that the ending `.dat` is not recognized (which is kinda annoying, as i use that ending for large input files that are supposed to be ignored by git and use `.csv` for shorter files that are supposed to be copied on the git).

But thats not the only issue. When i use the `MUSUN_100_events.csv` file that can be found in `examples/05-MUSUN`, i get: 
image

ManuelHu:
@EricMEsch the MUSUN reader should now work - I also added an automated test to run the example.

PR #243, opened 2025-01-25
TrackID in Germanium output scheme

### LINKED ISSUE #175: Add step and track information to output
adding `trackid` and `stepid` might be useful for clustering, to preseve energy deposition order

tdixon97:
- Adds optionally trackID and parentID to the Germanium output scheme.
- Also added some basic test in the `output` folder.
- added some docstrings for that output scheme while I was there.
This closes #175 
I would like to test this a bit better by compaing to the "track" output scheme.


include/RMGGermaniumOutputScheme.hh
```diff
@@ -26,20 +26,61 @@
...
-    void StoreEvent(const G4Event*) override;
-    bool ShouldDiscardEvent(const G4Event*) override;
+
+    /** @brief Store the information from the event, invoked in @c RMGEventAction::EndOfEventAction
+     */
+    void StoreEvent(const G4Event* event) override;
+
+    /** @brief Decide where to store the event, invoked in @c RMGEventAction::EndOfEventAction
+     *  @details @c true if the event should be discarded, else @c false .
+     *  The event is discard if there is not hit in Germanium of the energy range
```
ManuelHu, line 61:
```suggestion
     *  The event is discarded if there is no hit in Germanium or the energy range
```

include/RMGGermaniumOutputScheme.hh
```diff
@@ -26,20 +26,61 @@
...
+ *  - particle type,
+ *  - time,
+ *  - position,
+ *  - energy deposition,
+ *  - distance to detector surface.
+ *
+ * Optionally, the track ID and parent track ID can be stored as well as the
+ * detector ID for the single table output mode.
+ *
+ * In addition, this class is used for stacking tracks associated with optical
```
ManuelHu, line 43:
```suggestion
 * In addition, this class can be used for stacking tracks associated with optical
```

include/RMGGermaniumOutputScheme.hh
```diff
@@ -26,20 +26,61 @@
...
+    /** @brief Sets the names of the output columns, invoked in @c RMGRunAction::SetupAnalysisManager */
     void AssignOutputNames(G4AnalysisManager* ana_man) override;
-    void StoreEvent(const G4Event*) override;
-    bool ShouldDiscardEvent(const G4Event*) override;
+
+    /** @brief Store the information from the event, invoked in @c RMGEventAction::EndOfEventAction
+     */
+    void StoreEvent(const G4Event* event) override;
+
+    /** @brief Decide where to store the event, invoked in @c RMGEventAction::EndOfEventAction
```
ManuelHu, line 59:
```suggestion
    /** @brief Decide whether to store the event, invoked in @c RMGEventAction::EndOfEventAction
```

include/RMGGermaniumDetector.hh
```diff
@@ -25,6 +25,7 @@
 #include "G4VHit.hh"
 #include "G4VSensitiveDetector.hh"
 
+/** @brief Class to store hits in the Germanium detectors, wraps @c G4VHit */
```
ManuelHu, line 28:
```suggestion
/** @brief Class to store hits in the Germanium detectors, extends @c G4VHit */
```

ManuelHu (commented):
only some minor things in the docstring. Otherwise it is fine to be merged

tdixon97:
I am not sure how useful this will be for clustering though.
- I tried generating 1 MeV electrons and shooting them at a cylinder of Ge (along z with the Ge starting at -4 mm)
- I colored the steps by trackid, 
image
This seems to be working fine in some cases, when the secondary is an electron.
However, for other secondaries this is less useful
image
For example here we have secondary tracks but we have no information on where on the primary track they are produced so its not very useful for clustering.

EricMEsch:
> I guess something like an index of which step produced this track could be needed

If I recall correctly, that is not so trivial to implement. Because you can either get the secondaries produced within a step from the mother particle's point of view. But then that information would be part of the mother particle's hit event, and you would have to find a way to reasonably organize a bunch of secondaries in that hit table to be able to retrace their origins in post-processing. (Maybe some form of output to imitate a set containing the TrackID and the corresponding Step number? But I think it would be hard to make this output look comprehensible.)

And I think in the hit event of the secondary, the parent particle does not exist anymore. You only have its ParentID stored, but that's it. So you cannot access the Step number anymore. (I might be wrong, as I did not deeply investigate this, but from my superficial point of view, it looks like that.)


> For example here we have secondary tracks but we have no information on where on the primary track they are produced so its not very useful for clustering.

I would assume that is again because in the OutputScheme, only the `PostStepPoint` is saved. So the location of creation (which would be the `PreStepPoint`) is not. That's why it works fine in some cases, because, for e.g., electrons, the first `PostStepPoint` is not very far away from the original point of creation. So if you are only interested in the location of where the Track was created, that can be accessed by `G4Track->GetVertexPosition()` and if you are interested in the global time of when it was created you can try `G4Track->GetGlobalTime() - G4Track->GetLocalTime()`. (Or access this values in the `PreStepPoint` of the first step)

EricMEsch:
Wait a second, i just checked `RMGGermaniumDetector` and we are storing the `PreStepPoint`, not the `PostStepPoint`. I must have mixed that up, in the Optical detector we store the ``PostStepPoint`. But then i do not understand what might have caused Toby's second plot? Is the first step somehow not recognized inside the sensitive volume? Is the ParentID for the red secondary the cyan track?

ManuelHu:
> So you cannot access the Step number anymore

exactly: https://geant4-forum.web.cern.ch/t/get-g4track-by-trackid/1491

> But then i do not understand what might have caused Toby's second plot? Is the first step somehow not recognized inside the sensitive volume?

a guess: we only record steps with energy deposition > 0, so it _might_ actually miss steps that are only doing elastic scattering or creating secondaries (without _direct_ energy deposition)

tdixon97:
@EricMEsch Surely what is happening is that the Bremsttralhung process doesn't deposit any energy so the first step isn't recorded?
Maybe we want to revisit that choice, could be saving these few extra steps helps a lot in interpretation / clustering

tdixon97:
Yeah adding the 0 energy step helps. For example here is a bremsstralhung track.
image
You see the step with track ID = 2, when the bremsttralhung photon is produced. Then the next steps are from the secondary interactions.

PR #244, opened 2025-01-27
Add Particle filter stacking action option

EricMEsch:
Adresses #218.

Not tested so far. I did try the `make remage-doc-dump` as i add new macro commands here, but it did not update the `rmg-commands` file.

@tdixon97 I made a new PR instead of re-opening the old one, as i force pushed the changes back then into non-existence.

- [x] Compile tested
- [x] Write actual tests
- [x] Doxygen documentation


tdixon97:
@EricMEsch one thing I don't understand is why this would be a separate output scheme?

src/RMGParticleFilterOutputScheme.cc
```diff
@@ -0,0 +1,52 @@
...
+#include "G4EventManager.hh"
+#include "G4OpticalPhoton.hh"
+
+#include "RMGLog.hh"
+
+RMGParticleFilterOutputScheme::RMGParticleFilterOutputScheme() { this->DefineCommands(); }
+
+
+std::optional<G4ClassificationOfNewTrack> RMGParticleFilterOutputScheme::
+    StackingActionClassify(const G4Track* aTrack, int stage) {
```
tdixon97, line 30: Do we want to only stack particles after stage 1, eg. I am thinking about stacking nuclei only after they decay?

tdixon97:
I guess a simple test would be to run this with eg. alphas for some alpha decay and check that all the events are removed, and also a beta decay (checking nothing changes)?

EricMEsch:
> one thing I don't understand is why this would be a separate output scheme?

I mean it does not have to be. It is only utilizing the hook into the StackingAction for OutputSchemes. This could also be part of the StackingAction. I just thought it would be a better way of organizing to use this structure. If we are going to add multiple things to the StackingAction later on, we would be happy if they were not all hard coded into it. Also, this means that in case the OptionalOutputScheme is not activated by macro, this code will never be executed, which is a little bit faster i would assume.

> I guess a simple test would be to run this with eg. alphas for some alpha decay and check that all the events are removed, and also a beta decay (checking nothing changes)?

Could be worth a try, although i am not sure how Geant4 will react if you immediately kill the primary generated particle from the stack. In theory it should be fine i guess.

src/RMGParticleFilterOutputScheme.cc
EricMEsch: I am fairly certain the stages are something that is inherently handled by the user (us) and not G4? So far we run one stage, meaning everything until the Particle-Stack is empty.  Then we check if any OutputScheme requests a new stage. If one OutputScheme requests a new stage, all tracks that have been marked with `fWaiting` will be put on the stack again (which right now can only happen to `OpticalPhotons`). Otherwise they will be discarded. So i am not sure what exactly you refer to?

src/RMGParticleFilterOutputScheme.cc
```diff
@@ -0,0 +1,52 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+#include "RMGParticleFilterOutputScheme.hh"
+
+#include <set>
+
+#include "G4Event.hh"
+#include "G4EventManager.hh"
+#include "G4OpticalPhoton.hh"
```
ManuelHu, line 22:
```suggestion
```

include/RMGParticleFilterOutputScheme.hh
```diff
@@ -0,0 +1,49 @@
...
+
+#ifndef _RMG_PARTICLE_FILTER_OUTPUT_SCHEME_HH_
+#define _RMG_PARTICLE_FILTER_OUTPUT_SCHEME_HH_
+
+#include <optional>
+#include <set>
+
+#include "G4AnalysisManager.hh"
+#include "G4GenericMessenger.hh"
+#include "G4VUserEventInformation.hh"
```
ManuelHu, line 24:
```suggestion
#include "G4GenericMessenger.hh"
```

include/RMGParticleFilterOutputScheme.hh
```diff
@@ -0,0 +1,49 @@
...
+
+#include "RMGVOutputScheme.hh"
+
+class G4Event;
```
ManuelHu, line 28:
```suggestion
```

ManuelHu:
please also add to `remage-doc-dump.cc`, i.e. here: 

https://github.com/legend-exp/remage/blob/8ee3d58d4496ed4609ccdf607c1ebb7578971d1d/src/remage-doc-dump.cc#L48-L54

also, please rebase on main (on include #252) and run `make remage-doc-dump`

tdixon97:
@EricMEsch I wonder if it is possible to make this volume dependent or maybe G4Region dependent.
The reason is we are running some remage simulations for our detector at UCL, a screening like setup, but they are quite slow and I want to kill betas (only in the source not the detector of course)

EricMEsch:
It should certainly both be possible. For the volume dependence i guess a user would just have to specify the name of the physical volume.

I think a reasonable way to implement is:

- Delete the specified particle everywhere if no volume is specified
- As soon as at least one volume is specified only delete the particle in said volume.

The important thing to remember is, that this deletes on creation, so there is no way of knowing if the particle might have reached another volume.

tdixon97:
What about having either volumes to keep or volumes to kill ? I think the most common use case is kill everywhere apart from in a sensitive detector?

EricMEsch:
Should also be possible

The logic would then be

- Kill everywhere if nothing is specified
- If volume to keep is specified kill everywhere but said volume
- If volume to kill is specified kill only in that volume (ignore volumes to keep)

I guess i could add a warning if a volume to keep is specified in the same simulation with a volume to kill (as those are conflicting requests)

tdixon97:
Yeah alternatively we just have an option for the default and sensitive regions, these are already defined and I think then you can easily check which region a volume belongs to ? ie. [G4Region::BelongsTo](https://github.com/Geant4/geant4/blob/32390e802bd84a8b85ac02a90a3d31cf205dd5b7/source/geometry/management/include/G4Region.hh#L202C1-L202C56 )

I am not sure using the Regions is the right approach since its less easy to modify though..

EricMEsch:
I personally would prefer the volume by name option, but maybe that is just because i have not looked at the regions yet.
If that is easy i could also add that option. Although then the options might start to become overwhelming and confusing for the user.

EricMEsch:
@tdixon97 The option to add kill/keep volumes is now implemented. I am not 100% sure if the information about the physical volume actually exists at this point (as we catch the track before the tracking started), but i am pretty sure it does and we will quickly find out in testing.

The only "issue" now is that you can not specify a volume per particle. You specify volumes and particles and the filter is applied to all specified particles in all specified volumes.

It would be possible to use `std::map<int, std::vector<std::string>>` instead, but this would probably mean that i would have to add a custom messenger to allow the user to specify pdg codes togheter with a volume name.

tdixon97:
I think it sounds good, before doing very complicated  stuff we should consider what typical use cases are, probably killing beta or alpha anywhere not germanium.
I think if people want something very fancy they will prefer custom production cuts?

EricMEsch:
I have just done some tests (still need to add pipeline tests)

- Seems to work like expected, the only issue is if you filter the same particle as your primary. This results in an immediate segmentation fault (was kinda expecting G4 not to be happy with that). I might add an exception to that.
- If you specfiy a volume as both a keeper and a killer, an error will be shown but application will not end. If you continue anyways the volume will be treated as keeper. (Which is not equivalent to specifying the volume as keeper only, as the existence of a killer volume means that only particles in killer volumes will be deleted).
- The TrackID will still increase. E.g: An electron has 4 secondaries: e-,gamma,e-,gamma: If you run with tracking verbosity 2 it will still show you all 4 particles as being secondaries of the main particle. However the tracking then only shows TrackID 1 (e-) and TrackID 3 (e-). TrackID 2 and 4 is skipped (as the gammas are immediately killed before tracking).

tdixon97:
I think an exception for killing the primary is a very nice idea.
Overall this looks very nice. We could consider adding something to the validation suite about the effect of this.

EricMEsch:
This PR should be good to merge now i think?

include/RMGParticleFilterOutputScheme.hh
```diff
@@ -0,0 +1,83 @@
+// Copyright (C) 2022 Luigi Pertoldi <gipert@pm.me>
```
gipert, line 1: you own this code ;)

EricMEsch: :D i have not really cared much about the copyright stuff, i have mostly thought about this as enveloping the entire remage thingy and not that particular file.

tdixon97:
It would be nice to do some more testing before merging, ie compare some simulations with / without this option. And also maybe compare the run time.
I think a nice simulation would be gammas from an external source surrounded by air, and an Al shield (screening setup), we could simulate killing betas in the source and compare the spectrum?

EricMEsch:
I am not sure what exactly you mean? An external gamma source surrounded by air, what is the material/size of the source? Or why would there be betas in the source?

EricMEsch:
I think the feature itself is tested and works. It removes any specified particle in the specified volume before tracking.

The runtime and other comparison plots now really depend on the type of simulation you do. If you simulate muon showers and only kill neutrinos in a specified volume the runtime will not change much. If in your simulation gammas are 50% of the tracks and you kill all gammas then you are probably going to save 50% of runtime i would expect.

include/RMGParticleFilterOutputScheme.hh
gipert: I would prefer the actual original author being credited. @tdixon97 @ManuelHu @MoritzNeuberger Please also sweep through the remage source files and check this.

gipert:
What Toby meant is to have also something for the validation suite, e.g. a simple plot comparing energy spectra obtained with different kill-particle settings.

gipert:
Please rebase on main

gipert:
@EricMEsch I'd be happy to merge this and work on the validation plot, but I'd like to see it rebased in main first (there are some potentially impactful changes there). I cannot do it myself because you are working on your main branch (this is why you should never open pull requests from there but rather create a new branch), so only you can do it.

gipert:
@ManuelHu huh?

EricMEsch:
I will remember to create a new branch next time^^

What happened to the pre-commit? I tried pre-commit autoupdate, but still when i try to commit anything it literaly changes every file now. Is there something else i need to do to update pre-commit?

gipert:
What do you mean? pre-commit is happy with your PR here

src/RMGParticleFilterOutputScheme.cc
```diff
@@ -0,0 +1,104 @@
...
+  if (!fKeepVolumes.empty()) {
+    RMGLog::OutDev(RMGLog::fatal, "Conflicting requests for kill/keep volume in ParticleFilter. "
+                                  "Trying to assign kill-volume but a keep-volume already exists.");
+  }
+
+  fKillVolumes.insert(name);
+}
+
+std::optional<G4ClassificationOfNewTrack> RMGParticleFilterOutputScheme::
+    StackingActionClassify(const G4Track* aTrack, int stage) {
```
ManuelHu, line 46:
```suggestion
    StackingActionClassify(const G4Track* aTrack, int) {
```

EricMEsch:
> What do you mean? pre-commit is happy with your PR here

Yes, because i have only rebased and not made a new commit

i tried making a new commit changing the copyright author name, but then pre-commit goes crazy.

gipert:
Can you show here how you run pre-commit, your git diff and the output of pre-commit?

EricMEsch:
These are the changes i am doing:
image


So there are two scenarios, first if i just now install the remage-base_latest.sif fresh:
I run
```bash
pre-commit run --all-files
```
and get
image
This is probably due to something missing in the base container.

If i now instead go into my usual container (that i installed years ago and manage/update myself)

Everything passes except for clang

image

I feel like i have to manually update clang, i will try that

gipert:
So the problem with the clang hook is that it uses whatever Clang version you have locally. If you have a very recent one (e.g. on Arch Linux), it will format things differently. What I do is to always run pre-commit from within the container (or at least when C++ files are involved).

pre-commit is not installed in the remage-base container (I though I did, I'm fixing this right now). It seems like the pre-commit version you are using inside the container is too old, you need to find a way to update it (or wait until I update the container).

gipert:
>It sometimes happens that hooks need newer versions of pre-commit or identify, so I don't know if installing it in the base container makes too much sense; it might require too many rebuilds?

In my experience pre-commit is quite stable, so if I install it fresh from PyPI (working on it rn), I expect it to work fine for a long time.

ManuelHu:
But just to get this PR forward @EricMEsch: you can always use `git commit --no-verify [...]` to skip pre-commit checks; just to get the copyright change in right now.

EricMEsch:
Huh?

ManuelHu:
the CI failure can be ignored, just pip doing weird things

gipert:
I have updated remage-base, there's a newer pre-commit version now

Issue #190, opened 2024-12-09
Merge LH5 files from multiple threads


gipert:
Is it possible to handle this in pure C++ in a reasonable way? If not, we should provide some Python remage wrapper that performs the concatenation.

Yurivanderburg:
Not sure if this helps, but I found loading the lh5 files as awkward arrays and use ak.concatenate works very nicely. I can share some code if you want.

tdixon97:
This is exactly the approach we suggest people to use. 
However, the combination of the lh5 files directly in C++ / with an extra python script is a bit more tricky since you do not want to just concatenate but also sort by `g4_evtid` (since the files each contain a random set of `g4_evtids`). Then this also has to be done in a memory efficiency way (not reading the full data into memory), and it should be fast.

gipert:
We also want to do some merging in the simulation production workflow to avoid cluttering the filesystem with a huge amount of files.

gipert:
Now technically possible with #210.

ManuelHu:
Do we want to try to guess the output file names ourselves in python code, or do we want to have some way to communicate data from the `remage-cpp` process to the python wrapper?

gipert:
>the easies (infrastructure-wise) option would be to replicate the output file reanming scheme in python...

this is the way...

>there is also another problem: the number of threads also can be (indirectly) influenced via various env variables

then we'll have to understand how this works and implement protections in the python wrapper...

ManuelHu:
another complication: the outpur file name can be set in the macro (and not in the command line args). In this case we need remage reporting back the details to the python wrapper _anyway_ (unless we want to re-implement macro parsing in python ... 🫣)

tdixon97:
If remage now has a python wrapper we could potentially move away from geant4 macro files? Which are horrible.

gipert:
I am not sure we need to support naming output files with macro commands. Maybe we can just remove the functionality.

Toby, that sounds complicated. You still need a way to pass the configuration options to remage-cpp.

Manuel, should we think about moving all the code you wrote to make LH5 files to the Python wrapper at some point?

ManuelHu:
> should we think about moving all the code you wrote to make LH5 files to the Python wrapper at some point?

maybe, there were problems with h5py and the geant4-written files, the last time I tried.

Don't get me wrong, it would be very nice to get rid of this C++ code.

[...]

PR #246, opened 2025-01-28
add simple RMGIpc for remage-cpp -> python intercom

### LINKED ISSUE #261: Cleanup tmp files
Sometimes if you interrupt remage there are a bunch of temp files left on the disk, could they be deleted somehow?

ManuelHu:
**this is very much WIP and needs good testing before releasing**

this is a simple IPC mechanism. Its main features:
* one-way communication of unicode strings from the C++ side to python
* asynchronous messages do not get an acknowledgement back
   (this is meant for informational messages that the python code needs _after_ running remage-cpp, e.g. output file post-processing).
* synchronous messages require an acknowledgement signal from python. for this, the POSIX signal SIGUSR2 is used
   (this is meant for actions that python code should perform synchronously, e.g. working on input files)

by using this mechanism we can avoid re-implementing ugly logic in the python wrapper (e.g. guessing output file names, etc.) and get that data just form the C++ application.

this PR does not implement any actions on the python side yet. But with minor additions, this could be used for #190
* the biggest changes will be on the pydataobj side, as there is no way to call the functionality of `lh5concat` from python code...
* also, as @tdixon97 noted: `lh5concat` does currently load all files into memory, and does not use an iterator with smaller buffer size. This might not be what we want

fixes #261

### DIFF
--- a/include/RMGIpc.hh
+++ b/include/RMGIpc.hh
@@ -1,0 +1,48 @@
+// Copyright (C) 2025 Manuel Huber
+//
+// This program is free software: you can redistribute it and/or modify it under
+// the terms of the GNU Lesser General Public License as published by the Free
+// Software Foundation, either version 3 of the License, or (at your option) any
+// later version.
+//
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+// details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+#ifndef _RMG_IPC_HH_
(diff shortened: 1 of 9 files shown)


tdixon97:
@ManuelHu I guess it wouldnt be a big modification to make lh5concat iterate? Seems worthwhile also for other applications

gipert:
I think lh5concat changes shouldn't be hard to implement

ManuelHu:
this should now be mostly ready. I would like to implement any actual merging in a follow-up PR to keep apart the infrastructure work and the actual features.

tdixon97:
I don't understand what it is, but looks good

ManuelHu:
I think this should be ready right now, and already has some benefit for remage (without the `lh5concat` changes).

It also compiles & runs fine on macOS - even though we do not run most tests on macOS (because of missing HDF5 support), the python wrapper is executed and would log error messages if the ipc connection would not be working.

gipert:
good to merge from my side

PR #254, opened 2025-02-06
More messenger commands for physics lists

tdixon97:
I need to test it

### DIFF
--- a/include/RMGHardware.hh
+++ b/include/RMGHardware.hh
@@ -31,11 +31,13 @@
 #include "RMGNavigationTools.hh"
 #include "RMGVOutputScheme.hh"
 
+/** @brief Class to handle the detector geometry hardware, extends @c G4VUserDetectorConstruction . */
 class G4VPhysicalVolume;
 class RMGHardware : public G4VUserDetectorConstruction {
 
   public:
 
+    /** @brief Constructor based on calling the macro commands (with @c DefineCommands ). */
     RMGHardware();
     ~RMGHardware() = default;
 
@@ -44,23 +46,70 @@ class RMGHardware : public G4VUserDetectorConstruction {
     RMGHardware(RMGHardware&&) = delete;
     RMGHardware& operator=(RMGHardware&&) = delete;
 
+    /** @brief Construct the detector 
+     * 
+     *  @details All detector geometry must be based on GDML files, parsed with @c G4GDMLParser .
+     *  This method defines the geoemetry and checks for overlaps (if not disabled). It also
+     *  assigns physical volumes to Geant4 regions and sets user step limits.
+     *
+     * @returns The physical volume of the world.
+     * 
+     */
     G4VPhysicalVolume* Construct() override;
+
+    /** @brief Method to set the sensitive detectors. 
+     * 
+     * @details Setup thread-local data, called once for the master thread and once for each worker thread.
+     */
     void ConstructSDandField() override;
 
+    /** @brief Register a detector as being a remage sensitive detector.
+      * 
+      * @param type The type of detector.
+      * @param pv_name The name of the physical volume.
+      * @param uid A unique integer identifier for the sensitive volume.
+      * @param copy_nr The copy number for the physical volume.
+      * @param allow_uid_reuse Flag to allow a uid to be reused
+      */
     void RegisterDetector(RMGDetectorType type, const std::string& pv_name, int uid,
         int copy_nr = 0, bool allow_uid_reuse = false);
+
+    /** @brief Extract a map of the detector metadata, one element for each sensitive detector physical volume and copy_nr. */
     inline const auto& GetDetectorMetadataMap() { return fDetectorMetadata; }
+
+    /** @brief Extract the detector metdata for a given detector.
+      * 
+      * @param det the detector identifier, a pair of the physical volume name and the copy number.
+      */
     inline const auto& GetDetectorMetadata(const std::pair<std::string, int>& det) {
       return fDetectorMetadata.at(det);
     }
+
+    
     inline const auto& GetActiveDetectorList() { return fActiveDetectors; }
     [[nodiscard]] inline const auto& GetAllActiveOutputSchemes() { return fActiveOutputSchemes; }
 
+    /** @brief Add a GDML file to the geometry.
+      *
+      * @param filename file name to add. 
+      */
     inline void IncludeGDMLFile(std::string filename) { fGDMLFiles.emplace_back(filename); }
+
+    /** @brief Method to define geometry directly, the user must reimplement the base class method. */
     inline virtual G4VPhysicalVolume* DefineGeometry() { return nullptr; }
+
+    /** @brief Set the maximum step size for a certain physical volume.
+      * 
+      * @details This is used as a @c G4UserLimit to limit step sizes to being no larger than the chosen value.
+      * This requires the @c G4StepLimiter process to be activated in the physics list.
+      * 
+      * @param name The name of the physical volume to add step limits to.
+      * @param max_step The maximum step size.
+      */
     inline void SetMaxStepLimit(std::string name, double max_step) {
       fPhysVolStepLimits.insert_or_assign(name, max_step);
     }
+    
     inline void PrintListOfLogicalVolumes() { RMGNavigationTools::PrintListOfLogicalVolumes(); }
     inline void PrintListOfPhysicalVolumes() { RMGNavigationTools::PrintListOfPhysicalVolumes(); }
 
--- a/include/RMGHardware.hh
+++ b/include/RMGHardware.hh
@@ -28,6 +28,7 @@
 
 #include "RMGDetectorMetadata.hh"
 #include "RMGHardwareMessenger.hh"
+#include "RMGLog.hh"
 #include "RMGNavigationTools.hh"
 #include "RMGVOutputScheme.hh"
 
@@ -101,17 +102,16 @@ class RMGHardware : public G4VUserDetectorConstruction {
     /** @brief Method to define geometry directly, the user must reimplement the base class method. */
     inline virtual G4VPhysicalVolume* DefineGeometry() { return nullptr; }
 
-    /** @brief Set the maximum step size for a certain physical volume.
+    /** @brief Set the maximum step size.
      *
      * @details This is used as a @c G4UserLimit to limit step sizes to being no larger than the
      * chosen value. This requires the @c G4StepLimiter process to be activated in the physics list.
+     * This requires that @c fVolumeForStepLimit has been set.
      *
-     * @param name The name of the physical volume to add step limits to.
      * @param max_step The maximum step size.
      */
-    inline void SetMaxStepLimit(std::string name, double max_step) {
-      fPhysVolStepLimits.insert_or_assign(name, max_step);
-    }
+    inline void SetMaxStepLimit(double max_step);
+    inline void AddVolumeForStepLimits(std::string name) { fVolumeForStepLimit = name; };
 
     inline void PrintListOfLogicalVolumes() { RMGNavigationTools::PrintListOfLogicalVolumes(); }
     inline void PrintListOfPhysicalVolumes() { RMGNavigationTools::PrintListOfPhysicalVolumes(); }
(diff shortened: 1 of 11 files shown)


include/RMGHardware.hh
```diff
@@ -44,23 +46,70 @@ class RMGHardware : public G4VUserDetectorConstruction {
     RMGHardware(RMGHardware&&) = delete;
     RMGHardware& operator=(RMGHardware&&) = delete;
 
+    /** @brief Construct the detector
+     *
+     *  @details All detector geometry must be based on GDML files, parsed with @c G4GDMLParser .
```
ManuelHu, line 51: Not true. There always is the possibility to implement geometry in C++, as done in many of our examples.

also please state that this must not modify thread-local state, because it is only called once globally (I suppose?)

include/RMGHardware.hh
```diff
@@ -44,23 +46,70 @@ class RMGHardware : public G4VUserDetectorConstruction {
...
+     *  This method defines the geoemetry and checks for overlaps (if not disabled). It also
+     *  assigns physical volumes to Geant4 regions and sets user step limits.
+     *
+     * @returns The physical volume of the world.
+     *
```
ManuelHu, line 56:
```suggestion
```

include/RMGPhysics.hh
```diff
@@ -61,10 +90,12 @@ class RMGPhysics : public G4VModularPhysicsList {
...
 
-    // TODO: cut for optical photon?
-    struct StepCutStore {
-        StepCutStore() = default;
-        inline StepCutStore(double def_cut)
+    /** @brief Struct to hold the production cut values. */
+    struct ProdCutStore {
+        ProdCutStore() = default;
+
+        /** @brief Constructor setting the production cut for all particles to @c def_cut */
```
ManuelHu, line 97: mention "all particles" here, because that is quite misleading?

include/RMGPhysics.hh
```diff
@@ -76,19 +107,57 @@ class RMGPhysics : public G4VModularPhysicsList {
...
     };
 
+    /** @brief Sets the production cut values, and energy range.
+     *
+     * @details  Defines a set of production cuts for the default region
+     * and also for the G4Region "SensitiveRegion". This is created in RMGHardware, but this
+     * behavior might be changed in a derived class, so careful here!
+     * This function should be called after @c SetPhysicsRealm which is used to define the
+     * production cut values.
+     *
```
ManuelHu, line 117:
```suggestion
```

include/RMGHardwareMessenger.hh
```diff
@@ -18,22 +18,36 @@
 
 #include "G4UImessenger.hh"
 
+/** @brief Class for handling custom messenger commands. */
 class RMGHardware;
 class RMGHardwareMessenger : public G4UImessenger {
 
   public:
 
+    /** @brief Constructor based on a @c RMGHardware pointer. */
```
ManuelHu, line 27: I do not see the value in just having doc comments for the sake of having them... This commnet is just 100% the function signature.

src/RMGHardwareMessenger.cc
```diff
@@ -52,10 +66,23 @@ RMGHardwareMessenger::RMGHardwareMessenger(RMGHardware* hw) : fHardware(hw) {
...
 
-RMGHardwareMessenger::~RMGHardwareMessenger() { delete fRegisterCmd; }
 
-void RMGHardwareMessenger::SetNewValue(G4UIcommand* command, G4String newValues) {
-  if (command == fRegisterCmd) RegisterDetectorCmd(newValues);
+void RMGHardwareMessenger::DefineStepLimits() {
+
+  fStepLimitsCmd = new G4UIcommand("/RMG/Geometry/SetStepLimits", this);
+  fStepLimitsCmd->SetGuidance("set step limits for a sensitive detector");
+
```
ManuelHu, line 74:
```suggestion
```

tdixon97:
Any idea why my new macro command isn't picked up by `rmg-doc-dump`? I did run it

ManuelHu:
That would be a very curious case, as the CI also just runs `make remage-doc-dump`... (just running remage-doc-dump without make is not enough)

src/RMGHardwareMessenger.cc
```diff
@@ -52,10 +66,23 @@ RMGHardwareMessenger::RMGHardwareMessenger(RMGHardware* hw) : fHardware(hw) {
...
+
+  auto p_pv = new G4UIparameter("pv_name", 's', false);
+  p_pv->SetGuidance("Detector physical volume");
+  fStepLimitsCmd->SetParameter(p_pv);
+
+  auto step_limit = new G4UIparameter("step limit", 'd', true);
+  step_limit->SetGuidance("The maximum step limit");
+  step_limit->SetDefaultValue(1 * CLHEP::mm);
```
ManuelHu, line 82: there is also a way to have parameters with units, that would be nicer. Or, at least mention the unit this is in. Currently this just renders as "default value - 1" without any unit information in the docs.

tdixon97: I am working on it, but I had to change this into two macro commands

tdixon97: Now this is done

tdixon97:
The step limiting seems to be working, we could have a nicer interface (eg. specifying with regex) but I am not sure since I doubt its something we will use.

Here is an example with 10 um step limits:

image

gipert:
Seems good to me.

tdixon97:
I would wait until its more tested to merge.

src/RMGHardwareMessenger.cc
ManuelHu: You don't have to split it in two commands... I have some example code how to use G4UIcmdWithADoubleAndUnit with extra parameters:

```diff
diff --git a/include/RMGHardwareMessenger.hh b/include/RMGHardwareMessenger.hh
index 8ee2873..bcceed3 100644
--- a/include/RMGHardwareMessenger.hh
+++ b/include/RMGHardwareMessenger.hh
@@ -17,6 +17,7 @@
 #define _RMG_HARDWARE_MESSENGER_HH_
 
 #include "G4UImessenger.hh"
+#include "G4UIcmdWithADoubleAndUnit.hh"
 
 class RMGHardware;
 class RMGHardwareMessenger : public G4UImessenger {
@@ -31,7 +32,7 @@ class RMGHardwareMessenger : public G4UImessenger {
   private:
 
     RMGHardware* fHardware;
-    G4UIcommand* fRegisterCmd;
+    G4UIcmdWithADoubleAndUnit* fRegisterCmd;
 
     void RegisterDetectorCmd(const std::string& parameters);
 };
diff --git a/src/RMGHardwareMessenger.cc b/src/RMGHardwareMessenger.cc
index aa200af..1a4bdd8 100644
--- a/src/RMGHardwareMessenger.cc
+++ b/src/RMGHardwareMessenger.cc
@@ -17,14 +17,19 @@
 
 #include "G4Tokenizer.hh"
 #include "G4UIcommand.hh"
+#include "G4UIcmdWithADoubleAndUnit.hh"
 
 #include "RMGHardware.hh"
 #include "RMGTools.hh"
 
 RMGHardwareMessenger::RMGHardwareMessenger(RMGHardware* hw) : fHardware(hw) {
-  fRegisterCmd = new G4UIcommand("/RMG/Geometry/RegisterDetector", this);
+  fRegisterCmd = new G4UIcmdWithADoubleAndUnit("/RMG/Geometry/RegisterDetector", this);
   fRegisterCmd->SetGuidance("register a sensitive detector");
 
+  fRegisterCmd->SetParameterName("test", false);
+  fRegisterCmd->SetDefaultValue(5);
+  fRegisterCmd->SetUnitCategory("Length");
+
   auto p_type = new G4UIparameter("type", 's', false);
   p_type->SetParameterCandidates(RMGTools::GetCandidates<RMGDetectorType>().c_str());
   p_type->SetGuidance("Detector type");
@@ -59,9 +64,15 @@ void RMGHardwareMessenger::SetNewValue(G4UIcommand* command, G4String newValues)
 }
 
 void RMGHardwareMessenger::RegisterDetectorCmd(const std::string& parameters) {
+  RMGLog::Out(RMGLog::error, "params ", parameters);
   G4Tokenizer next(parameters);
 
+  auto x = next();
+  auto y = next();
+  auto num = G4UIcmdWithADoubleAndUnit::GetNewDoubleValue((x + " " + y).c_str());
+
   auto type_str = next();
+  RMGLog::Out(RMGLog::fatal, " num ", num);
   auto type = RMGTools::ToEnum<RMGDetectorType>(std::string(type_str), "detector type");
   auto pv_name = next();
   const int uid = std::stoi(next());
```

tdixon97: I am not sure I understand this code

tdixon97:
Everything should be ready now, but regarding the results one thing is confusing me:
If I simulate with small step limits and keeping 0 energy hits I still get some isolated steps not part of a track eg.
image
image

Sometimes I see multiple steps in the same gamma track but not so often, does anyone understand why. I was expecting a step when the gamma is created and one when it interacts? Maybe @EricMEsch  has some idea

tdixon97:
image
Step limits seem to work fine (this is 50 um) although I am not sure about the line with steps of 40 um..

src/RMGPhysics.cc
```diff
@@ -448,8 +451,51 @@ void RMGPhysics::SetPhysicsRealm(PhysicsRealm realm) {
...
+        " while the realm has already been explicitly set, this will override the production cuts");
+  }
+
+  RMGLog::OutFormat(RMGLog::summary,
+      "Setting user defined production cuts for sensitive region to {} mm", cut / u::mm);
+
+  fProdCutsSensitive = ProdCutStore(G4VUserPhysicsList::defaultCutValue);
+  fProdCutsSensitive.gamma = cut;
+  fProdCutsSensitive.electron = cut;
+  fProdCutsSensitive.positron = cut;
```
ManuelHu, line 470: I guess it is intended that this omits proton and generic_ion? If so, this shuld be documented (both for the function in an doxygen string, and for the messenger)...

ManuelHu (commented):
the updated command for step sizes in volumes look good. there are just some remnants of intermediate solutions that you had implemented...

src/RMGPhysics.cc
tdixon97: Yeah I guess production cuts are not usually used for alpha and ions since most processes do not produce them as secondary (only radioactive decays)

tdixon97:
Should all be ready to merge now then @ManuelHu

src/RMGPhysics.cc
```diff
@@ -533,15 +533,17 @@ void RMGPhysics::DefineCommands() {
 
   fMessenger
       ->DeclareMethodWithUnit("DefaultProductionCut", "mm", &RMGPhysics::SetDefaultProductionCut)
-      .SetGuidance("Set simulation production cuts, for default region. Note: this overrides the "
-                   "physics realm")
+      .SetGuidance("Set simulation production cuts, for default region for proton, position, "
+                   "electron and gamma. Notes: this overrides the "
+                   "values from the physics realm. This does not apply to proton or generic ions.")
```
ManuelHu, line 538: this now contradicts itself w.r.t. protons? :-)
```suggestion
      .SetGuidance("Set simulation production cuts, for default region for electrons, positions, "
                   "and gammas. Notes: this overrides the values from the physics realm. This does "
                   "not apply to protons, alphas or generic ions.")
```

src/RMGPhysics.cc
```diff
@@ -533,15 +533,17 @@ void RMGPhysics::DefineCommands() {
...
       .SetParameterName("cut", false)
       .SetStates(G4State_PreInit, G4State_Idle);
 
   fMessenger
       ->DeclareMethodWithUnit("SensitiveProductionCut", "mm", &RMGPhysics::SetSensitiveProductionCut)
-      .SetGuidance("Set simulation production cuts, for default region. Note: this overrides the "
-                   "physics realm")
+      .SetGuidance("Set simulation production cuts, for default region for proton, position, "
+                   "electron and gamma. Notes: this overrides the "
+                   "values from the physics realm. This does not apply to proton or generic ions.")
```
ManuelHu, line 546: same here, also was still mentioning "default region".

```suggestion
      .SetGuidance("Set simulation production cuts, for sensitive region for electrons, positions, "
                   "and gammas. Notes: this overrides the values from the physics realm. This does "
                   "not apply to protons, alphas or generic ions.")
```

ManuelHu:
updated command docs still missing (`make remage-doc-dump`), then I would merge it

gipert:
@ManuelHu can we have the doc check as a separate CI step, running after one of the other steps? Would make it simpler to understand what's wrong.

ManuelHu:
that would require the second CI job to run the compilation again, no? (it already is a seperate _step_, so I assume you mean a seperate _job_?)

gipert:
Yes, I meant as a separate job. But it looks like we would have to use the cache to let it start from where the first job stopped, annoying. Nevermind. Maybe we could add some more helpful message, for when the doc-dump step fails.

ManuelHu:
Another thing I just looked into: it is quite non-trivial to add changes to a pull request from inside an action run. So automating this further would be quite complex

but I can work on a better error message

gipert:
Toby could you tidy up the commit history? It's quite ugly...

tdixon97:
Is it ok now @gipert ?

gipert:
I had to re-run remage-doc-dump in [bd14996](https://github.com/legend-exp/remage/commit/bd149962f1beaa48329d443d8c898bf7088c2d88), somehow...

PR #259, opened 2025-02-09
ci: better check for rmg command docs

ManuelHu:
this is just a cosmetic test to see if we can make the check for up-to-date command docs more nice-looking. no logic changes


gipert:
@ManuelHu this is good to merge from my point of view. My preference is for the CI to fail fast if the docs are not all right.

PR #264, opened 2025-02-12
Add commands to control which step point is stored for Germanium

tdixon97:
This adds also the post-step position (and distance to surf) to the RMGGermaniumHit. It then adds a macro command to allow the user to chose which position to store in the output.
@EricMEsch it includes some minor refactor of your code so maybe you can see if its ok?

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -80,19 +80,21 @@ set_target_properties(
 
 # install section
 
+set(INSTALL_VENV_DIR ${CMAKE_INSTALL_PREFIX}/share/remage_venv)
+
 # install the package into the install prefix together with a virtualenv
 # in this way remage should always run isolated in its environment.
 # we want to use the cpp_config for the install area
 add_custom_command(
   OUTPUT ${CMAKE_INSTALL_PREFIX}/bin/remage
-  COMMAND ${Python3_EXECUTABLE} -m venv ${CMAKE_INSTALL_PREFIX}
-  COMMAND ${CMAKE_INSTALL_PREFIX}/bin/python -m pip -q install --no-warn-script-location --upgrade
-          pip
-  COMMAND ${CMAKE_INSTALL_PREFIX}/bin/python -m pip -q install --no-warn-script-location --upgrade
-          uv
+  COMMAND ${Python3_EXECUTABLE} -m venv ${INSTALL_VENV_DIR}
+  COMMAND ${INSTALL_VENV_DIR}/bin/python -m pip -q install --no-warn-script-location --upgrade pip
+  COMMAND ${INSTALL_VENV_DIR}/bin/python -m pip -q install --no-warn-script-location --upgrade uv
   COMMAND cp ${CMAKE_CURRENT_BINARY_DIR}/cpp_config.install.py
           ${CMAKE_CURRENT_SOURCE_DIR}/remage/cpp_config.py
-  COMMAND ${CMAKE_INSTALL_PREFIX}/bin/python -m uv -q pip install --reinstall ${CMAKE_SOURCE_DIR}
+  COMMAND ${INSTALL_VENV_DIR}/bin/python -m uv -q pip install --reinstall ${CMAKE_SOURCE_DIR}
+  COMMAND ln -s ${INSTALL_VENV_DIR}/bin/remage ${CMAKE_INSTALL_PREFIX}/bin/remage
+  WORKING_DIRECTORY ${CMAKE_INSTALL_PREFIX}
   COMMENT "Installing remage Python wrapper")
 
 add_custom_target(install-remage-cli DEPENDS ${CMAKE_INSTALL_PREFIX}/bin/remage)


ManuelHu:
How much computational time do we waste by calculating 3 distances, and only using one?

Apart from that, it looks mostly fine.

src/RMGGermaniumDetector.cc
```diff
@@ -123,33 +150,23 @@ bool RMGGermaniumDetector::ProcessHits(G4Step* step, G4TouchableHistory* /*histo
...
-    const auto sample_solid = sample_physical->GetLogicalVolume()->GetSolid();
-    const double sample_dist = sample_solid->DistanceToIn(sample_point);
-    if (sample_dist < dist) { dist = sample_dist; }
-  }
 
-  hit->distance_to_surface = dist;
+  // get various distances
+  hit->distance_to_surface_prestep = DistanceToSurface(sv, lv, tf, position_prestep);
+  hit->distance_to_surface_poststep = DistanceToSurface(sv, lv, tf, position_poststep);
+  hit->distance_to_surface_average = DistanceToSurface(sv, lv, tf, position_average);
```
EricMEsch, line 169: It is not guaranted that all of these steps are inside of the volume. I am not sure what would happen then. (I think if i remember correctly G4 might return kInfinity?)

The PostStepPoint might be outside of the volume at boundaries (PreStepPoint is inside, PostStepPoint is not at boundaries when it leaves the volume. The Average can also be outside for other reasons)

The `ProcessHits()` function checks if the physical volume of the PreStepPoint is registered as a detector. It does not check for the other Points.

src/RMGGermaniumOutputScheme.cc
```diff
@@ -170,13 +171,29 @@ void RMGGermaniumOutputScheme::StoreEvent(const G4Event* event) {
...
+                                   ? hit->global_position_prestep
+                               : (fPositionMode == RMGGermaniumOutputScheme::PositionMode::kPostStep)
+                                   ? hit->global_position_poststep
+                                   : hit->global_position_average;
+
+      double distance = (fPositionMode == RMGGermaniumOutputScheme::PositionMode::kPreStep)
+                            ? hit->distance_to_surface_prestep
+                        : (fPositionMode == RMGGermaniumOutputScheme::PositionMode::kPostStep)
+                            ? hit->distance_to_surface_poststep
+                            : hit->distance_to_surface_average;
```
EricMEsch, line 187: I am not sure about coding style etc.

But wouldnt it be more intuitive (and mabye quicker, although i guess that effect is negligible) to just check
```c++
if(...)
{
Do both assign operations here
}
```
My point is not about the use of the `?` operator, but rather doing the same operations once for the position and once for the distance, although it only has to be done once

include/RMGGermaniumDetector.hh
```diff
@@ -84,6 +90,16 @@ class RMGGermaniumDetector : public G4VSensitiveDetector {
...
+    /** @brief Compute the distance from the point to the surface of the physical volume.
+     * @details Checks distance to surfaces of mother volume.
+     * @param sv The solid to find the distance to.
+     * @param lv The logical volume instance, used to check for mother volume.
+     * @param tf The transformation to global coordinates
+     * @param position The position to evaluate the distance for.
+     */
+    double DistanceToSurface(const G4VSolid* sv, const G4LogicalVolume* lv,
+        const G4AffineTransform tf, const G4ThreeVector& position);
+
```
EricMEsch, line 102: I guess this is mostly cosmetical, but you could just hand over the physical volume and the position to this function, You can then derive the lv, sv and transform within the function.

But i guess it does not really matter

src/RMGGermaniumOutputScheme.cc
ManuelHu: I would also prefer using if statements here. Just for clarity, though - that statement is really not nice to read

src/RMGGermaniumDetector.cc
tdixon97: This is slightly tricky since `ProcessHits` decides if a hit is in the Ge detector before the output scheme decides which distance to store.
Are we ok with removing hits where the post-step point is outside the detector? 
This will happen if a track starts on the surface and goes out.

tdixon97:
@EricMEsch does this look ok now?

src/RMGGermaniumDetector.cc
tdixon97: I think I agree with you @ManuelHu. Safety in `geant4` should prevent this..
So then I would propose to check that the pre-step and post-step point are in the same volume and if not discard the step?
Does this work for all particles, even gammas?

EricMEsch: Yes, you are right, it should not be an issue as long as we do not start applying electromagnetic fields in our detector (at which point the steps might become curved i think).

I am not entirely sure if we want to remove steps which post-step point is outside of the volume. Geant4 handels steps leaving the boundary in the way that the PreStepPoint is inside the volume and the PostStepPoint, while the position is exactly on the surface, is defined as outside of the volume. But as the position will be exactly on the surface this might still be fine with the DistanceToOut() function? I am not sure, but i would first do some testing before we start removing steps, that clearly should have deposited all energy inside of the volume.

tdixon97: > Yes, you are right, it should not be an issue as long as we do not start applying electromagnetic fields in our detector (at which point the steps might become curved i think).

Is that true, I thought steps were always linear but can be drawn curved in visualisation (the `rich` mode of track rendering)?

src/RMGGermaniumDetector.cc
```diff
@@ -112,8 +133,37 @@ bool RMGGermaniumDetector::ProcessHits(G4Step* step, G4TouchableHistory* /*histo
...
+  const auto prestep = step->GetPreStepPoint();
+  const auto position_prestep = prestep->GetPosition();
+
+  const auto poststep = step->GetPostStepPoint();
+  const auto position_poststep = poststep->GetPosition();
+  const auto position_average = (position_prestep + position_poststep) / 2.;
+
+  // check containment of pre and post-step point
+  auto prestep_inside = this->CheckStepPointContainment(prestep);
+  auto poststep_inside = this->CheckStepPointContainment(poststep);
```
EricMEsch, line 157: I am not sure we want to check the Containment of the PostStep. Steps that leave the volume will have their entire step within the volume and the PostStep will be on the surface but `pv = step_point->GetTouchableHandle()->GetVolume();` will be the other volume as defined by G4.

I would rather just make a sanity check to see that the `DistanceToOut(PostStepPoint)` provides reasonable results. This could just be a seperate test in the pipeline like my distance_to_surface test but with the PostStepPoint selected

include/RMGGermaniumOutputScheme.hh
```diff
@@ -84,13 +91,17 @@ class RMGGermaniumOutputScheme : public RMGVOutputScheme {
...
+    inline void SetPositionMode(PositionMode mode) { fPositionMode = mode; }
+
   protected:
 
     [[nodiscard]] inline std::string GetNtuplenameFlat() const override { return "germanium"; }
 
   private:
 
     RMGGermaniumDetectorHitsCollection* GetHitColl(const G4Event*);
+    void SetPositionModeString(std::string mode);
```
EricMEsch, line 104: Have you tried function-overloading? I am not sure if this is possible in combination with the messenger? But if it works i think it would be nicer to just overload the `SetPositionMode()` function.

src/RMGGermaniumDetector.cc
EricMEsch: When a particle enters the volume the prestep will be not in the volume and the poststep will be exactly on the surface but `pv = step_point->GetTouchableHandle()->GetVolume();` will return the volume. But from a physical perspective it is fine to discard this step, as the entire steplength is outside of the volume.

For the particle leaving the entire steplength would be inside the volume, so we do not want to discard it i think?

src/RMGGermaniumDetector.cc
EricMEsch: I am not an expert on this, i know seeing curved steps in some simulations other people once showed me at VIEWS, so i thought it had to, but that was just the visualization. There is the `G4ChordFinder` to handle such things, but i am not exactly sure what the G4 navigation is doing then.

src/RMGGermaniumDetector.cc
EricMEsch: That should be handled fine. Every step will be stopped at the boundary when it enters a new volume.

The step where a particle enters a new volume will be handled such that the PostStep, which is at the boundary, is assigned to the volume the particle is entering. From this point on tracking works normally again.

This means the next step the prestep will be the new volume and the poststep will be as well (untill the particle leaves the volume, then the poststep will be the new volume again)

EricMEsch: I think that it is also typically not possible for a particle to change direction on the boundary? As the step is limited by the boundary and not a physics process, the particle will continue its direction into the new volume and can not be immediately backscattered (at least not untill the next step in the new volume).

But for this reason there are some specialized backscattering solutions and physics processes that deal with this issue for low energy particles. But this is where it gets complicated again and i am not too sure what happens again.

PR #266, opened 2025-02-13
Fix Python wrapper install

gipert:
I realized that the installed remage wrapper is completely borked: it is set to use the Python interpreter from the virtualenv in the build area. I didn't realize this because we exclusively test remage in the build area (bad idea).

- [x] fix removal of src/build dirs in Dockerfile
- [x] fix python wrapper installation by installing a virtualenv in the install area
- [x] test installed executable in CI

### DIFF
--- a/.dockerhub/Dockerfile
+++ b/.dockerhub/Dockerfile
@@ -33,6 +33,7 @@ RUN mkdir -p src build /opt/remage && \
     else \
         ctest --output-on-failure --label-exclude 'vis'; \
     fi && \
+    cd .. && \
     rm -rf build src
 
 ENV PATH="/opt/remage/bin:$PATH" \
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-
-mkdir build
-cd build || exit 1
-cmake ..
-make
-make install
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -145,9 +145,4 @@ jobs:
           cd build
           ctest --output-on-failure --label-exclude 'extra|vis'
 
-      - name: Minimally test installed executable
-        run: |
-          rm -rf build
-          PATH="$PWD/local/bin:$PATH" DYLD_LIBRARY_PATH="$PWD/local/lib:$DYLD_LIBRARY_PATH" remage --version
-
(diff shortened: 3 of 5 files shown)


ManuelHu:
`/usr/bin` is not writable on the macOS runner - we are not in a container there.

ManuelHu:
missing mkdir for the `local` dir :-)

gipert:
I think CMake should make it...

gipert:
Oh great, aren't the G4 libs supposed to be already visible?

ManuelHu:
conda apparantly does not set a DYLD_LIBRARY_PATH and advises against it: https://docs.conda.io/projects/conda/en/latest/user-guide/troubleshooting.html

but I don't know enough about macOS and conda on macOS how to fix this.

gipert:
Also people complaining here: https://github.com/conda/conda/issues/12800

gipert:
I would skip this test on macos for now.

ManuelHu:
Yes, apparently conda really wants people to do it "their way" and use a tool to fixup the rpath in all conda-installed executables/libraries.

This works fine for using conda-installed libraries with conda-installed executables.

but this obviously stops working when you try to compile stuff inside that env

ManuelHu:
probably setting DYLD_LIBRARY_PATH to include `$CONDA_PREFIX/lib` might work? I also needed to set a similar env var for CMAKE:..

gipert:
Feel free to push a commit if you want to give it a try, or just merge...

ManuelHu:
a seperate PR will be easier to test

ManuelHu:
@gipert another thing that came to my mind. We should advise people to _never_ install remage to `/usr`, and ideally also not into `/usr/local`.

Especially the first will quite certainly mess up any linux system as this would let `uv` work on the system python.

But also an installation into `/usr/local` can be very problematic, as the venv contains a lot of other binary scripts (and not only the `remage.cli` entry point...)

gipert:
Yes, I agree. We could print a warning and mention it in the installation guide.

EricMEsch:
But this only means the `usr/local` folder or also sub-folders? As i have installed remage to `usr/local/remage-install` so far

ManuelHu:
Yes, only the prefixes `/usr/local` or  `/usr/` without subdirectories would be affected (or, generally: any prefix where there is another python installation located). So anything below that would be safe.

PR #269, opened 2025-02-14
output: swap ntuple group name and ntuple name

### LINKED ISSUE #250: Swap `stp` and `ch{}` in the otput table names
I realized that in the LEGEND data we have `ch00000/raw` while remage writes `stp/ch00000`. I like the second option more because then `stp` can be an `LGDO.Struct`, but it's better to keep consistency.

ManuelHu:
@gipert have fun with the tests 😉

fixes #250


gipert:
💥

ManuelHu:
as described in Luigi's issue, it will change the output structure from `stp/<detector_name>` to `<detector_name>/stp`

Luigi will fix the tests and continue this PR until it's ready, he just asked me to quickly implement the core change.

gipert:
We are proposing to go for `stp/<detector_name>` in LEGEND data now, so I'll close (for now).

PR #274, opened 2025-02-20
python: install venv into subdirectory

### LINKED ISSUE #272: python wrapper installation is having funny side-effects
This has been meantioned by @EricMEsch in a recent discussion.

currently, the python wrapper is installed as a full python venv into the system $PATH. This is not only problematic as it might mangle with the system python installation (mitigation in PR #270), but also has another side effect.

The system path does not just contain the remage python wrapped binary, but a large number of other binaries:

* all entry points of packages, e.g. `lh5ls`, `pint-convert`...
* `python`...
* `pip`...
* `uv`...
* `activate`

This is highly problematic for any use outside of the official tagged container images, as those binaries might take precendence over system-provided binaries (i.e. you cannot run anything in the system python installation any more!), or over other virtual environments if the user is not careful.

ManuelHu:
fixes #272

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -80,19 +80,21 @@ set_target_properties(
 
 # install section
 
+set(INSTALL_VENV_DIR ${CMAKE_INSTALL_PREFIX}/share/remage_venv)
+
 # install the package into the install prefix together with a virtualenv
 # in this way remage should always run isolated in its environment.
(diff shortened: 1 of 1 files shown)


gipert:
Makes sense, if you have verified that it does what we discussed in #272.

gipert (approved):

ManuelHu:
works locally for me, at least

PR #281, opened 2025-03-08
update fmt and fix some compat issues, get fmt from the web with CMake

gipert:


ManuelHu:
this breaks `clang-tidy`. it now reports errors (and tries to fix them...) on the fmt code

and `FetchContent_Declare(... SYSTEM)` to mark the library as a system dependency is not yet supported on ubuntu 22.04

gipert:
oof. the first we can probably fix but the second? i would have really liked to stop having a copy of fmt in the source tree. it could also be used for magic_enum

PR #282, opened 2025-03-08
Enable more compiler warnings, turn them into errors 😈

gipert:


gipert:
Also this fixes all warning with my gcc version (14.2.1) 🎉

PR #283, opened 2025-03-09
Get CLI11 and magic_enum with CMake's FetchContent

gipert:
This does not work, but I don't understand why.


gipert:
Thanks a lot Manuel

PR #290, opened 2025-03-17
support wrapper-only args and macro substitutions

gipert:
We can now substitute variables in template macro files (e.g. `$ENERGY` -> `1000`) at runtime. This is going very helpful in the dataflow and for studying systematics.
```
remage --substitute-in-macro ENERGY=1000 -- macro.mac
```
where `macro.mac`:
```
/RMG/Geometry/RegisterDetector Germanium germanium 001

/run/initialize

/RMG/Generator/Confine Volume
/RMG/Generator/Confinement/Geometrical/AddSolid Sphere
/RMG/Generator/Confinement/Geometrical/Sphere/OuterRadius 1

/RMG/Generator/Confinement/Geometrical/CenterPositionX 0
/RMG/Generator/Confinement/Geometrical/CenterPositionY 0
/RMG/Generator/Confinement/Geometrical/CenterPositionZ 0

/RMG/Generator/Select GPS
/gps/particle e-
/gps/ang/type iso
/gps/energy $ENERGY keV

/run/beamOn 1000
```

I had to figure out how to support args parsing in the Python wrapper and forwarding to `remage-cpp`. I had to change quite some code, sorry.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -3,8 +3,15 @@
 
 set(_r ${PROJECT_SOURCE_DIR})
 
-set(PYTHON_SOURCES ${_r}/cmake/cpp_config.py.in ${_r}/python/remage/__init__.py
-                   ${_r}/python/remage/cli.py ${_r}/python/remage/ipc.py ${_r}/pyproject.toml)
+# cmake-format: off
+set(PYTHON_SOURCES
+    ${_r}/pyproject.toml
+    ${_r}/cmake/cpp_config.py.in
+    ${_r}/python/remage/__init__.py
+    ${_r}/python/remage/cli.py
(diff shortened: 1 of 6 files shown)


ManuelHu:
just thinking about it, why not using geant4 aliases and exposing a nice way to set aliases on CLI?

They would also work in macros called with `/control/execute` and in interactive mode - unlike this implementation... The syntax is also similar:

```
# need to construct this from CLI and inject it (in RMGManager?)
/control/alias ENERGY 1000

#  ... the actual user macro
/gps/energy {ENERGY} keV
```

gipert:
Superseded by #291. I will not delete the branch, there's maybe some useful code for the future.

PR #293, opened 2025-03-17
Validation: electron ranges in germanium

gipert:
As expected, the simulated range is longer than the CSDA range (increasing with energy) due to bremsstrahlung emission (my guess) --- see new plots in the tarball. Next steps could be disabling the bremsstrahlung process or "correct" the CSDA range curve.


gipert:
@tdixon97 I guess we expect there rare random failures

https://github.com/legend-exp/remage/actions/runs/13910267114/job/38923636871#step:7:182

Issue #289, opened 2025-03-17
Add `cmake-lint` to pre-commit config


gipert:
https://cmake-format.readthedocs.io/en/latest/cmake-lint.html

I haven't tested it but it could be useful.

EricMEsch:
Was this fixed with the PR? If not just re-open

PR #294, opened 2025-03-17
pre-commit: try to add cmake-lint

ManuelHu:
see #289 , let's see what happens


gipert:
I was expecting more useful linting... I actually disagree with the naming convention

ManuelHu:
we can set our own ones

ManuelHu:
only some things remaining (mostly long lines, and some local var casing)

ManuelHu:
do we actually need `cmake/modules/DependencyGraph.cmake`? I just tried it and I do not fid the output particularly useful? (`make dep-graph-remage`)

gipert:
Is it so ugly? It's not so necessary

ManuelHu:
image

this is the dep graph of the `remage` target - the only thing from us is the one node are ~5 nodes at the top,, the rest is just Geant4, Qt, ROOT & X11 libraries?

gipert:
Amazing @tdixon97 we should show this to the collaboration

tdixon97:
make the same for mage

PR #300, opened 2025-03-18
User manual

gipert:
- [x] figure out how to cross-reference API docs and RMG macro commands

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -8,13 +8,14 @@ dependencies:
   - python=3.12
   - pip
   - breathe
+  - docutils>=0.21
   - doxygen
-  - exhale
+  - exhale>=0.3.7
   - furo
   - graphviz
   - myst-parser
   - setuptools-scm
-  - sphinx
+  - sphinx>=8.2
   - sphinx-copybutton
   - sphinx-togglebutton
   - sphinx-inline-tabs
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -9,7 +9,6 @@ dependencies:
   - breathe
   - docutils>=0.21
   - doxygen
-  - exhale>=0.3.7
   - furo
   - graphviz
   - myst-parser
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -6,7 +6,6 @@ channels:
 
 dependencies:
   - python=3.12
-  - pip
   - breathe
   - docutils>=0.21
   - doxygen
@@ -19,3 +18,6 @@ dependencies:
   - sphinx-copybutton
   - sphinx-togglebutton
   - sphinx-inline-tabs
+  - pip:
+    dependencies:
+      - exhale>=0.3.7
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -17,6 +17,4 @@ dependencies:
   - sphinx-copybutton
   - sphinx-togglebutton
   - sphinx-inline-tabs
-  - pip:
-    dependencies:
-      - exhale>=0.3.7
+  - pip: ["exhale>=0.3.7"]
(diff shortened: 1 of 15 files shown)


gipert:
What are all these CI false positives now

ManuelHu:
`[bool RMGVertexConfinement::SampleableObject::Sample(G4ThreeVector&, size_t, bool, size_t&) const] Generated vertex not inside sampling volumes (forced containment check)`

not sure though why this is happening more often recently (is a _volume_ containment check even always accurate for a _surface_ sampler?)

tdixon97:
I am also confused, ill look into it

tdixon97:
In general the confinement test should still work, although it seems to rarely fail (possibly some numerical issue).
One idea could be to change `IsInside` from:

``` c++
if (navigator->LocateGlobalPointAndSetup(vertex) == this->physical_volume) return true;
```
To something like:

```c++
auto local_pos = this->rotation.inverse() * vertex - this->translation;
this->physical_volume->GetLogicalVolume()->GetSolid()->Inside(local_pos) == EInside::kInside
```
Then we can explicity check `EInside::kSurface` for the surface generators?

ManuelHu:
> figure out how to cross-reference API docs and RMG macro commands

it should be fairly straighforward to add a custom labels to the auto-generated docs? Then they could be referenced the "usual" rst way?
The other way round is slightly harder, depends if we want to have/tolerate rst/md in the geant4 GUI help text etc.

ManuelHu:
might be worth a try

tdixon97:
Maybe we should merge this? I can help filling in some parts

gipert:
sure if you don't have major comments about the structure

tdixon97:
Id move input event vertices to vertex confinement / event generators respectively. Otherwise it looks good

tdixon97:
I imagine it as a subsection of event generators and of vertex confinement

gipert:
I would have expected this `<project:rmg-commands#rmg-manager-interactive>` to work but it does not:

```
/home/gipert/sw/src/legend/remage/build/docs/index.md:5: WARNING: local id not found in doc 'rmg-commands': 'rmg-manager-interactive' [myst.xref_missing]
```

ManuelHu:
does not work on rtd?

docs/manual/output.md
```diff
@@ -1,7 +1,27 @@
...
+
+_remage_ supports all output formats supported by
+[`G4AnalysisManager`](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/Analysis/managers.html)
+(HDF5, ROOT, CSV, XML), plus
+[LH5](https://legend-exp.github.io/legend-data-format-specs/dev/hdf5). The file
+type to use is selected by the specified output file name (`.h5`, `.root`,
+`.csv`, `.xml`, `.lh5`).
+
+:::{note}
+HDF5 and ROOT output formats require Geant4 to be explicitly compiled with
```
ManuelHu, line 22:
```suggestion
LH5, HDF5 and ROOT output formats require Geant4 to be explicitly compiled with
```

gipert:
>does not work on rtd?

great

gipert:
Works locally. I wonder what could be the problem...

ManuelHu:
nice. this install issue has been reported upstream _yesterday_, and is a conda-specific one. https://github.com/conda-forge/exhale-feedstock/issues

gipert:
nice find. so maybe we need to play with the mambaforge version?

docs/environment.yml
```diff
@@ -8,13 +8,14 @@ dependencies:
   - python=3.12
   - pip
```
ManuelHu, line 9:
```suggestion
  - pip
    dependencies:
      - exhale>=0.3.7
```

PR #312, opened 2025-03-31
Adding observable plots to the validation report

tdixon97:
Includes some plots of:
- total energy
- active energy
- r90
- range
For 1 MeV surface and bulk electrons. It might be cumbersome to add more stuff eg. scans in energy, we might want to think of a more elegant solution. Eg. interactive figures with sliders


docs/validation/observables-ge.md
```diff
@@ -24,3 +25,267 @@ This geometry is shown below, for gamma particles generated in the source.
...
+### Bulk electrons
+
+We generate 1 MeV electrons in the bulk of the Germanium detector varying
+the step limits and production cuts. This is relevant both for the simulation of neutrinoless double beta decay
+but also any gamma's depositing energy in the HPGe. This is since a change in the fraction of electrons with fully reconstructed
+energy will change the peak-to-continuum ratio for the gamma simulation.
+
+#### Total energy
+
+We extract the total energy deposited in the Germanium detector, the energy spectra is plotted below:
```
EricMEsch, line 155: > the energy spectra

spectrum? It is a single spectrum below :D

docs/validation/observables-ge.md
```diff
@@ -24,3 +25,267 @@ This geometry is shown below, for gamma particles generated in the source.
...
+
+```{figure} ./_img/observables/beta-observables.bulk-total-energy.spec.output.png
+:width: 800px
+Spectrum of total energy in Germanium for 1 MeV electrons generated uniformly in the volume. We compare the spectrum without
+step limits to that with 10 um and 200 um limit. The residual shows the difference between the two extreme cases.
+```
+
+&nbsp;
+
+To quantify this we count the fraction of events without 1 keV of the full energy peak.
```
EricMEsch, line 165: > without 1 keV

within 1 keV?

docs/validation/observables-ge.md
```diff
@@ -24,3 +25,267 @@ This geometry is shown below, for gamma particles generated in the source.
...
+```{figure} ./_img/observables/beta-observables.bulk-r90.spec.output.png
+:width: 800px
+ Spectrum of r90 in Germanium for 1 MeV electrons generated on the surface. We compare the spectrum without
+step limits to that with 10 um and 200 um limit. The residual shows the difference between the two extreme cases.
+```
+
+&nbsp;
+
+To quantify the effect of the step limits on our observable of interest (the fraction of multisite events), we compute
+the fraction of events with > 1 mm r90.
```
EricMEsch, line 216: But this does not really represent or show the shift in r90 we see in the previous plot.

docs/validation/observables-ge.md
```diff
@@ -24,3 +25,267 @@ This geometry is shown below, for gamma particles generated in the source.
...
+
+```{figure} ./_img/observables/beta-observables.surf-active-energy.eff.output.png
+:width: 800px
+Fraction of generated electrons with more than 300 keV active energy, as a function of the step limits.
+```
+
+&nbsp;
+
+Again it is possible you may see differences relating to the step limits. However, it should be kept in mind that for simulations of beta's, which is a continuous
+spectrum this detector response function will be conolved with the spectrum shape minimising the effect.
```
EricMEsch, line 271: I mean yes, but this only really hides, but does not fix the effect, that we may have slight deviation regarding the deposited energy on an event by event basis

docs/validation/observables-ge.md
```diff
@@ -24,3 +25,267 @@ This geometry is shown below, for gamma particles generated in the source.
...
+
+```{figure} ./_img/observables/beta-observables.surf-max-z.spec.output.png
+:width: 800px
+Distribution of the range of the electrons in Germanium for 1 MeV electrons generated on the surface. We compare the spectrum without
+step limits to that with 10 um and 200 um limit. The residual shows the difference between the two extreme cases.
+```
+
+&nbsp;
+
+To quantify this we extract the fraction of events with a range > 1 mm.
```
EricMEsch, line 286: It really seems like the effect is maximal in the 0.75-1mm range? For > 1mm there is not really much statistics left and the fluctuation is really high

docs/validation/observables-ge.md
tdixon97: I use plural since there are different spectra for different step limits.

docs/validation/observables-ge.md
tdixon97: Yeah I agree, but it is more relevant for whats important to us, the fraction of multi vs single site events. 
We could also use maybe the mean r90 as another parameter

docs/validation/observables-ge.md
tdixon97: Sure but the question is always whether an effect is large enough to care about in a given application. It is hard for us to run a simulation with enough primaries that after the convolution with the beta shape you still see an effect. But maybe we could comment better

docs/validation/observables-ge.md
tdixon97: When I defined this range I was not really aiming to increase the level of variation with cuts but rather to think about what`s a meaningful observable. But maybe also eg. the mean range is useful? I would say people get enough information from the plot

docs/validation/observables-ge.md
EricMEsch: But then it would be "the energy spectra *are.
.." xD

Issue #313, opened 2025-03-31
Output format: step grouping and windowing


tdixon97:
Currently we write the outputs as a flat table. This was initially motivated by the limitations of the HDF5 writing in Geant4.
However, with the python wrapper (#210 and others), it is now feasible to do something nicer.

I think already doing step grouping (see reboost docs for info), in the python wrapper would have a number of advantages. I would then propose an output format like:
```
g4_evtid:      [[1,1,1], [2,2,2], ...]
time:          [[0,1,2],[], []]
edep:          [[100,200,300],[],...]
```
where every field is a `VectorOfVectors`:

Advantages:
-----------------
- easier for users (one less step)
- lossless, if you want another grouping just flatten the output and redo the grouping,
- ensures that different programs (reboost, PSS etc.) all use the same indexing
- generally the step grouping has a minimal effect and doesn't need to be changed often

However, this would be a rather major change, better to do it now!

other things to think about:
- do we want to also group together different channels at this step, might involve padding some files with NaNs or Zeros (but would compress well)?

EricMEsch:
> do we want to also group together different channels at this step

How exactly do you mean this grouping? If we do it this way the first entry for every channel would be corresponding to the first event (and can have any size as it is a vector, so there would be no padding required here?)

tdixon97:
> But i think there still should do the macro option to not do it.
> 
> In the end it is not providing any real value except moving a step from the Post-processing into the Simulation chain?

Hmm, the point is that for some uses (eg. PSS) the analysis is not done in python and so we have to be very very careful to use the same indexing. If instead we enforcing the indexing from the beginning I think its alot safer.

I am not sure we want to propogate a large number of options (it will get confusing), so I would propose the jagged structure as the only one. If a user wants a flat table they just have to flatten it.

tdixon97:
> > do we want to also group together different channels at this step
> 
> How exactly do you mean this grouping? If we do it this way the first entry for every channel would be corresponding to the first event (and can have any size as it is a vector, so there would be no padding required here?)

I just meant do we want the same indexing between channels, which would make the data already look a bit "event" structured (we do this with the data) and make some things easier later

SalehGiovanna:
I think having a "global trigger" structure which mimics the data structure could be a good idea, especially if we want to go in the direction of processing simulated data with the same tools as real data.

For the PSS side, I am not sure whether the current code can handle [NaN] entries, but I can modify it so as to produce a flat waveform when the energy deposit is NaN

[...]

PR #314, opened 2025-04-01
python: add more infrastructure for post-processing

### LINKED ISSUE #190: Merge LH5 files from multiple threads

ManuelHu:
see #190 @tdixon97 

this deliberately fails pre-commit and some mt tests, this will both be fixed when the actual merging is patched in

### DIFF
--- a/python/remage/cli.py
+++ b/python/remage/cli.py
@@ -186,8 +186,32 @@ def remage_run(
     }
     logger.setLevel(levels_rmg_to_py[log_level])
 
-    # TODO: further post-processing
-    _output_files = ipc_info.get("output")
+    # output post-processing (merging multiple LH5 files)
+    output_files = ipc_info.get("output")
+    main_output_file = ipc_info.get_single("output_main", None)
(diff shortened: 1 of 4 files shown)


tdixon97:
Thanks @ManuelHu I am currently testing the performance of lh5concat, then I guess it will be ok to merge from the lgdo side and we can move on with this

tdixon97:
@ManuelHu do you have some idea how we should fit #313 into this. Ideally any basic post-processing will happen before dumping the data to disk

ManuelHu:
> Ideally any basic post-processing will happen before dumping the data to disk

this will not be possible as Geant4 will always write the data to disk in per-thread files. We can only implement transformations of the kind `n_thread -> m` whereas typically `m = 1` .

> Could be an alternative of the `lh5concat` command?

yes, we could technically add a `reboost` command instead of `lh5concat` here. the rest of the infrastructure should be the same.

PR #320, opened 2025-04-05
gdml: add command to load detectors from GDML

### LINKED ISSUE #319: uid in macro and GDML
Currently the mapping from uid --> name is provided both in the geant4 macro and the GDML file. This gives a danger of inconstiency in the two.
We should move away from the `detectors.mac` files and just grab the uids from GDML? Or atleast check the mappings are the same

ManuelHu:
fixes #319

### DIFF
--- a/docs/manual/geometry.md
+++ b/docs/manual/geometry.md
@@ -5,5 +5,33 @@
 - remage design principle and pointers to pyg4ometry and friends
 - how to input a GDML geometry
 - navigating the geometry tree within remage
-- registering sensitive detectors
   :::
+
+## Registering sensitive detectors
+
+It is important to register all sensitive detector volumes that should store
+particle interactions into output files. In remage, there are multiple ways
+to register such a sensitive volume. Internally, each sensitive detector has
+an id number (`uid`, whereas the `u` stands for unique. However, an id can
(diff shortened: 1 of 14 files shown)


ManuelHu:
this should now be ready to review. As outlined in #319 I cannot think about any warnings/errors (about name/id duplication, ...) that we still need to implement for this feature to work as expected?

Any name-reuse or id-reuse will lead to an error or warning, as the detectors from GDML should be loaded _last_.

note: adding two detectors to one physvol is just emitting a warning (since forever?) this logic is not changed
https://github.com/legend-exp/remage/blob/c25c21ebd9411124efb3809b28f3df59e525ed69/src/RMGHardware.cc#L231-L235

gipert:
Error if no mapping is available in GDML?

ManuelHu:
Should we extend this to register some detector types only? I think it might be useful to be able to do `/RMG/Geometry/RegisterDetectorsFromGDML Optical` or so?

EricMEsch:
Yes, i think that is a good idea

gipert:
Do you want to dump some bullet points in the docs / manual about this? I don't want you to spend time on this but then someone else can expand them in the near future.

PR #325, opened 2025-04-15
python: add kwargs to remage_run

ManuelHu:
finally found 5 minutes to adjust the tests ;-) @tdixon97

### DIFF
--- a/python/remage/cli.py
+++ b/python/remage/cli.py
@@ -172,7 +172,7 @@ def remage_run(
         args.append(f"--gdml-files={gdmls}")
 
     if output is not None:
-        args.append(f"--output={output}")
+        args.append(f"--output-file={output}")
 
     args.append(f"--threads={threads}")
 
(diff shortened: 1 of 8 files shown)


python/remage/cli.py
```diff
@@ -145,6 +164,57 @@ def remage_run(
...
     """
+    args = []
+    if not isinstance(gdmls, str):
+        for gdml in gdmls:
+            args.append(f"--gdml-files={gdml}")
+    else:
+        args.append(f"--gdml-files={gdmls}")
+
+    if output is not None:
+        args.append(f"--output={output}")
```
EricMEsch, line 175: I dont understand, why would `--output` work? The keywords are `-o` or `--output-file` according to the .cpp main.

It is the same in the basic tutorial it says `--output`, but that does not work for me?

ManuelHu: yes exactly 😅

PR #331, opened 2025-04-17
preclustering of the remage outputs and refactor of output schemes / detectors

tdixon97:
Changed this PR into being for the pre-clustering, still very WIP.
 - [x] basic code to write out clustered steps
 - [x] cluster between tracks
 - [x] different threshold for surface and bulk
 - [x] recompute dist to surface
 - [x] messenger commands to control it all
 - [x] Implement for the Scintillator output scheme
 - [x] documentation
 - [ ] some validation suite tests - **can be maybe skipped for now**


tdixon97:
I added to this PR a macro command for forcing to apply cuts, relevant for discussions of low energy secondary from compt and phot processes. I will also add here some plots showing the effect of this.

tdixon97:
The macro command has the expected effect
image
However, on more reflection I am not sure this is a good idea, since for our (fairly high) production cuts a significant fraction of Compton scatters will be suppressed. Of course `Geant4` conserves energy and gives this energy to the parent gamma track, but this would make it a bit harder to keep track of where the energy deposit is.
So I would not use this option and instead perform a clustering in the output scheme...

EricMEsch:
I agree. I would also not use this as default

tdixon97:
Most tests pass - which is promising, I am just checking those that didn't..

gipert:
One preliminary note: I would like to see this code very well documented, both code comments and docs.

tdixon97:
Sure, of course I will do it.

tdixon97:
This now works for clustering within a track. We could already start testing this, since it should be the same as the method in reboost.

include/RMGGermaniumDetector.hh
```diff
@@ -95,7 +97,7 @@ class RMGGermaniumDetector : public G4VSensitiveDetector {
      * @param pv The physical volume to find the distance to.
      * @param position The position to evaluate the distance for.
      */
-    double DistanceToSurface(const G4VPhysicalVolume* pv, const G4ThreeVector& position);
+    static double DistanceToSurface(const G4VPhysicalVolume* pv, const G4ThreeVector& position);
```
gipert, line 100: does this need to be a class method?

include/RMGPhysics.hh
```diff
@@ -129,6 +129,15 @@ class RMGPhysics : public G4VModularPhysicsList {
...
     void SetPhysicsRealmString(std::string realm);
 
+    /** @brief Set production cuts to be enabled for all EM processes.
+     *
+     *  @details Some EM physics lists only apply production cut thresholds to certain processes,
+     *  e.g. those with UV divergence. This can be used to instead force the cuts to be applied
+     *  for all processes. This can be useful to control the output file size.
+     *
+     */
+    void SetEMApplyCutsAll();
```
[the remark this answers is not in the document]
tdixon97: I am probably going to remove this.

include/RMGGermaniumDetector.hh
gipert: let's try to move code that does not logically belong to this class to another namespace. I think clustering algos could just go to their own file. See e.g. RMGGeneratorUtils

tdixon97: But Distance to surface does logically belong to the Germanium detector class?

src/RMGGermaniumDetector.cc
```diff
@@ -184,9 +188,9 @@ bool RMGGermaniumDetector::ProcessHits(G4Step* step, G4TouchableHistory* /*histo
   hit->parent_track_id = step->GetTrack()->GetParentID();
 
   // get various distances
-  hit->distance_to_surface_prestep = DistanceToSurface(pv, position_prestep);
-  hit->distance_to_surface_poststep = DistanceToSurface(pv, position_poststep);
-  hit->distance_to_surface_average = DistanceToSurface(pv, position_average);
+  hit->distance_to_surface_prestep = DistanceToSurface(hit->physical_volume, position_prestep);
+  hit->distance_to_surface_poststep = DistanceToSurface(hit->physical_volume, position_poststep);
+  hit->distance_to_surface_average = DistanceToSurface(hit->physical_volume, position_average);
```
EricMEsch, line 193: What is the reason for going to `hit->physical_volume`? I mean they hsould be the same, but staying with pv should be cleaner (and little bit faster)?

include/RMGGermaniumDetector.hh
gipert: why? doesn't it work for any pv?

tdixon97: Yeah, that's why I made it static since it doesn't use any member variables in the end

src/RMGGermaniumOutputScheme.cc
```diff
@@ -143,8 +144,123 @@ bool RMGGermaniumOutputScheme::ShouldDiscardEvent(const G4Event* event) {
...
+  hit->distance_to_surface_prestep = hits.front()->distance_to_surface_prestep;
+  hit->distance_to_surface_poststep = hits.back()->distance_to_surface_poststep;
+  hit->distance_to_surface_average =
+      RMGGermaniumDetector::DistanceToSurface(hits.back()->physical_volume,
+          hit->global_position_average);
+
+  return hit;
+}
+
+RMGGermaniumDetectorHitsCollection* RMGGermaniumOutputScheme::PreClusterHits(
```
gipert, line 185: I would like to see more comments here

gipert:
I'm ok with merging this functionality for now and implement inter-track clustering in a future PR. I would however think better about the code design. Also, do we really need the "pre" before "clustering"?

tdixon97:
I think its better to implement the full clustering, it shouldn't take too long

src/RMGGermaniumOutputScheme.cc
```diff
@@ -143,8 +144,123 @@ bool RMGGermaniumOutputScheme::ShouldDiscardEvent(const G4Event* event) {
...
+  hit->global_position_poststep = hits.back()->global_position_poststep;
+
+  // issue: this could be outside the volume!
+  hit->global_position_average = (hit->global_position_prestep + hit->global_position_poststep) / 2.;
+
+  hit->distance_to_surface_prestep = hits.front()->distance_to_surface_prestep;
+  hit->distance_to_surface_poststep = hits.back()->distance_to_surface_poststep;
+  hit->distance_to_surface_average =
+      RMGGermaniumDetector::DistanceToSurface(hits.back()->physical_volume,
+          hit->global_position_average);
```
EricMEsch, line 180: Add a check if the average Point is in the volume? Either by calling `IsInside` from the solid of the physical volume (a little bit more effort) or check if `hit->distance_to_surface_average` has a reasonable value. I think if it is not Inside this should have a wierd value.

src/RMGGermaniumOutputScheme.cc
```diff
@@ -143,8 +144,123 @@ bool RMGGermaniumOutputScheme::ShouldDiscardEvent(const G4Event* event) {
...
+
+  for (auto hit : *hits->GetVector()) {
+
+    if (!hit) continue;
+
+    // within track clustering
+    bool start_new_cluster =
+        (prev_hit == nullptr) or (hit->track_id != prev_hit->track_id) or
+        (hit->detector_uid != prev_hit->detector_uid) or
+        (std::abs(hit->global_time - prev_hit->global_time) > fClusterTimeThreshold);
```
EricMEsch, line 204: This heavily relies on the order of the initial `RMGGermaniumDetectorHitsCollection`. It should work, because typically Geant4 will finish tracking a particle before tracking secondaries, but i think there is an option to change that. In that case the `trackID`s would not be ordered and this would not cluster at all?

It should definitly be documented how exactly this works.

src/RMGGermaniumOutputScheme.cc
```diff
@@ -143,8 +144,123 @@ bool RMGGermaniumOutputScheme::ShouldDiscardEvent(const G4Event* event) {
...
+
+    // check distances
+    if (!start_new_cluster) {
+      bool is_surface = hit->distance_to_surface_average < fSurfaceThickness;
+      bool is_surface_prev = prev_hit->distance_to_surface_average < fSurfaceThickness;
+
+
+      start_new_cluster =
+          (is_surface and not is_surface_prev) or (not is_surface and is_surface_prev);
```
EricMEsch, line 213: `start_new_cluster = is_surface != is_surface_prev`

src/RMGGermaniumOutputScheme.cc
```diff
@@ -143,8 +144,123 @@ bool RMGGermaniumOutputScheme::ShouldDiscardEvent(const G4Event* event) {
...
+
+      // get the right distance to pre-cluster
+      double threshold = is_surface ? fClusterDistanceSurface : fClusterDistance;
+
+      start_new_cluster =
+          start_new_cluster ||
+          (hit->global_position_average - prev_hit->global_position_average).mag() > threshold;
+    }
```
EricMEsch, line 221: I am not the biggest fan of this double assignment.
Maybe call it `bool surface_transition = is_surface != is_surface_prev`
And then do `start_new_cluster = surface_transition || ...`

src/RMGGermaniumOutputScheme.cc
```diff
@@ -163,8 +279,6 @@ void RMGGermaniumOutputScheme::StoreEvent(const G4Event* event) {
 
       if (!hit or (hit->energy_deposition == 0 and this->fDiscardZeroEnergyHits)) continue;
 
-      hit->Print();
```
EricMEsch, line 166: But then you can not get this print anymore if you do it without preclustering?

As you moved the print into your preclustering code.

include/RMGGermaniumOutputScheme.hh
```diff
@@ -118,7 +136,15 @@ class RMGGermaniumOutputScheme : public RMGVOutputScheme {
     bool fStoreSinglePrecisionPosition = false;
 
     bool fStoreTrackID = false;
+    bool fPreClusterHits = true;
+
+    // clustering pars
+    double fClusterTimeThreshold = 10 * CLHEP::us;
+    double fClusterDistance = 10 * CLHEP::um;
+    double fClusterDistanceSurface = 1 * CLHEP::um;
+    double fSurfaceThickness = 2 * CLHEP::mm;
```
EricMEsch, line 145: Do we want macro commands for the time as well?

src/RMGGermaniumOutputScheme.cc
```diff
@@ -143,8 +144,123 @@ bool RMGGermaniumOutputScheme::ShouldDiscardEvent(const G4Event* event) {
...
+RMGGermaniumDetectorHitsCollection* RMGGermaniumOutputScheme::PreClusterHits(
+    const RMGGermaniumDetectorHitsCollection* hits) {
+
+  // create a container for the output hits
+  auto out = new RMGGermaniumDetectorHitsCollection();
+
+  std::vector<std::vector<RMGGermaniumDetectorHit*>> hits_vector;
+
+  // keep track of the current cluster
+  RMGGermaniumDetectorHit* prev_hit = nullptr;
```
EricMEsch, line 194: The name "prev_hit" is missleading. This is not the previous hit, it is the first hit of the cluster.

include/RMGGermaniumDetector.hh
EricMEsch: But thats what Luigi says, that he thinks this should be moved in a separate utils class. Then GermaniumDetector would call this utils class, but also other stuff can call it.

src/RMGGermaniumOutputScheme.cc
EricMEsch: Also even within track this clustering is not perfect. As You are starting a new cluster as soon as a hit is `>threshold` away from the initial point. This means if the particle is backscattered and gets closer again it will be at the same position but in a new cluster.

But i guess for our use case this is good enough?

tdixon97: I think its important to remember this is "**pre**"-clustering. The idea is not to have something optimal but just to remove totally unneeded information. Any more serious clustering should be done in post-processing.

src/RMGGermaniumOutputScheme.cc
tdixon97: Yeah that is a tricky point, you have to rely on the ordering to some extent. We should think of the right checks to make

src/RMGGermaniumOutputScheme.cc
tdixon97: I thought about that, its true the point could be outside in some cases, but then I am not sure what we want to do?

EricMEsch: At least we should give some kind of warning or anything. Not just pretend like it didnt happen

src/RMGGermaniumOutputScheme.cc
```diff
@@ -143,8 +144,123 @@ bool RMGGermaniumOutputScheme::ShouldDiscardEvent(const G4Event* event) {
...
+  hit->particle_type = hits.front()->particle_type;
+  hit->track_id = hits.front()->track_id;
+  hit->parent_track_id = hits.front()->parent_track_id;
+
+  // time from first hit
+  hit->global_time = hits.front()->global_time;
+
+  // positions from first and last step
+  hit->global_position_prestep = hits.front()->global_position_prestep;
+  hit->global_position_poststep = hits.back()->global_position_poststep;
```
EricMEsch, line 171: This also is pretty arbitrary. It is neither given that the first step is the one the most in one direction. Neither is it given that the last step is the furthest away from it... 

It is not like it matters, as we are clustering anyways and it should not matter where we put the cluster in the end... But this seems like a lot of logic that just does random things.

Why do we even need a prestep or poststep point for the cluster? Can we just place it at one point?

src/RMGGermaniumOutputScheme.cc
EricMEsch: This should also only be able to occur for a cluster at the Surface. Right now you are clustering with reduced range, but still clustering at the surface. Maybe we should just not cluster at the surface? Then this should not be an issue

src/RMGGermaniumOutputScheme.cc
tdixon97: Well we are clustering together steps in the same track, which are ordered. So I dont agree it is "arbitrary" . We need the pre-step and post-step point incase the user wanted this in their output. It would be possible to get rid of that but this goes against the philosophy of the pre-clustering.

tdixon97: It is certainly an option to remove fields, like the "pre" / "post" step which are not necessarily so meaningful. But I don't see the advantage. All we are doing is merging a bunch of steps. Logically there is still a pre-step and a post-step?

EricMEsch: What i mean is that you could have something like this (and i dont even think that this is rare, we saw plenty of steps like this at the end of a trajectory)
grafik

In this case the "PreStep" and "PostStep" do not really have a meaning... 
I do not like having an option that sounds relevant or meaningfull when it is not...

tdixon97:
That could anyway happen within one geant4 step

src/RMGGermaniumOutputScheme.cc
tdixon97: Thats also possible with the current implementation, just set the surface threshold to 0. But even at the surface I would do some level of pre-clustering, eg. 100 nm steps still don't mean anything.

src/RMGGermaniumOutputScheme.cc
EricMEsch: I mean yes, it should not be of a concern. But that is what i mean that these "PreStep" or "PostStep" fields also are not really of a concern...

include/RMGGermaniumDetector.hh
tdixon97: I am not sure about making new classes just to hold functions, I would rather it not be a method of any class

ManuelHu: we already have `RMGNavigationTools`, maybe it fits there?

gipert: >we already have RMGNavigationTools, maybe it fits there?

well it's not navigation. I would make a new file

>I am not sure about making new classes just to hold functions, I would rather it not be a method of any class

should be functions in a namespace. see RMGNavigationTools

src/RMGGermaniumOutputScheme.cc
tdixon97: I think this is now partly fixed since I explicity loop first over the trackid, but still we assume some ordering within each track. Have a look @EricMEsch

src/RMGGermaniumOutputScheme.cc
tdixon97: I added some, still we need more docstring details

include/RMGGermaniumDetector.hh
tdixon97: It is sort of navigation

tdixon97:
@gipert / @EricMEsch / @ManuelHu this is now updated to do some between track clustering, with one approach, have a look!

include/RMGGermaniumDetector.hh
gipert: what's your problem with well structured code

include/RMGGermaniumOutputScheme.hh
```diff
@@ -72,6 +72,53 @@ class RMGGermaniumOutputScheme : public RMGVOutputScheme {
...
+     * fClusterTimeThreshold ,
+     * - the distance between the step and the first of the cluster is not above @c fClusterDistance
+     * (for the bulk) or
+     * @c fClusterSurfaceDistance for the surface.
+     * - the hits in each cluster are then combined into one effective step with @c AverageHits
+     *
+     * @returns a collection of hits after pre-clustering.
+     */
+    RMGGermaniumDetectorHitsCollection* PreClusterHits(
+        const RMGGermaniumDetectorHitsCollection* hits);
```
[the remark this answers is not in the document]
gipert: i don't think we need this overloading feature... advanced users can overload the upstream method that processes hits

tdixon97:
Started making some sanity checks on the track combination. This shows the energy per track with and without the track combine, with a threshold of 10 keV. So still sometimes you have tracks < 10 keV since there was no close neighbour track found.
image

I think otherwise it makes sense:
- events at around 9 keV originate from X-rays so the largest track is the 9 keV one,
- similar for above 10 keV
- around 7 keV must be some Auger electrons? and similar at low energy?

Overall, it seems to be working and cleans up considerably the output

include/RMGOutputTools.hh
```diff
@@ -0,0 +1,97 @@
...
+      double cluster_distance_surface;
+      double cluster_time_threshold;
+  };
+
+  /** @brief Compute the distance from the point to the surface of the physical volume.
+   * @details Checks distance to surfaces of mother volume.
+   * @param pv The physical volume to find the distance to.
+   * @param position The position to evaluate the distance for.
+   */
+  double DistanceToSurface(const G4VPhysicalVolume* pv, const G4ThreeVector& position);
```
gipert, line 41: If a function is not part of a class, we use the `distance_to_surface` casing

include/RMGOutputTools.hh
```diff
@@ -0,0 +1,97 @@
...
+   * within the cluster distance of this track. If so they are combined for further
+   * pre-clustering. In the case multiple nearby tracks are found the highest
+   * energy one is used.
+   *
+   * @returns A map of steps after combining low energy tracks.
+   */
+  std::map<int, std::vector<RMGDetectorHit*>> CombineLowEnergyElectronTracks(std::map<int,
+                                                                                 std::vector<RMGDetectorHit*>>
+                                                                                 hits_map,
+      ClusterPars cluster_pars);
```
gipert, line 91: wtf is this indentation

tdixon97: whatever your pre-commit thing did

tdixon97: it really insists on that choice

gipert: might fix it by sending the first arg to a new line

src/RMGPhysics.cc
```diff
@@ -112,6 +113,10 @@ void RMGPhysics::SetStoreICLevelData(bool store) {
 
 ////////////////////////////////////////////////////////////////////////////////////////////
 
+void RMGPhysics::SetEMApplyCutsAll() { G4EmParameters::Instance()->SetApplyCuts(true); }
```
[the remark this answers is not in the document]
tdixon97: we could remove it? Not sure, is it nice for some application?

include/RMGOutputTools.hh
tdixon97: it didnt

src/RMGPhysics.cc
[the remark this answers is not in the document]
gipert: bah, maybe better to stick to the philosophy of implementing only what is really needed...

include/RMGOutputTools.hh
```diff
@@ -0,0 +1,97 @@
...
+   * total energy in each electron track.
+   *  If a track is below a certain threshold then the code searches through the
+   * other tracks to see if there is one where the first pre-step point is
+   * within the cluster distance of this track. If so they are combined for further
+   * pre-clustering. In the case multiple nearby tracks are found the highest
+   * energy one is used.
+   *
+   * @returns A map of steps after combining low energy tracks.
+   */
+  std::map<int, std::vector<RMGDetectorHit*>> combine_low_energy_tracks(std::map<int,
```
gipert, line 88: @ManuelHu is there a clang-format option to make it always put the first argument on a new line, like for the Python linting?

tdixon97:
This now includes the macro commands to make a "light" scintillator output.
The only missing piece is the possibility to give gamma track energy to the daughters and not write them out. Is it something we really want? @gipert ?

tdixon97:
We should also decide on the default parameters: Id propose:

- **positions**: float32 for scintillator and float64 for germanium
- **distance**: float32 and on for Germanium by default
- **step point**: average for both
- **velocity**: float32 and off by default
- **trackid**: saved by default (both scint. and ge.)
- **germanium** clustering: 10um (bulk), 1um (surface), 2mm surface region
- **scintillator clustering**: 100 um
- **track merging**: 10 keV threshold for both

This would be a bit of a change of philosophy compared to before, only saving what's necessary unless the user insists.

include/RMGOutputTools.hh
ManuelHu: `AllowAllArgumentsOnNextLine: true`

but we cannot do this in this PR. unlike the name, it does not just **allow** that , but **force** that style. Changing this and running clang-format changes (nearly) **every file**

ManuelHu: we can make a seperate PR after this one

gipert: agreed

gipert:
>The only missing piece is the possibility to give gamma track energy to the daughters and not write them out. Is it something we really want? @gipert ?

honestly, yes? how much would we save?

gipert:
Toby I think the proposed parameters are ok, and we can always adjust them in a separate PR

tdixon97:
> > The only missing piece is the possibility to give gamma track energy to the daughters and not write them out. Is it something we really want? @gipert ?
> 
> honestly, yes? how much would we save?

A fraction, the gammas are ~50% of the tracks and <20% of the steps (after basic clustering)

gipert:
@EricMEsch @ManuelHu please review

gipert:
Then I think we should

gipert:
I think we can set it to be default already

gipert:
maybe it would be worth to already drop some docs in the manual...

tdixon97:
> Then I think we should

@gipert  I now added this feature

src/RMGOutputTools.cc
```diff
@@ -110,11 +110,53 @@ RMGDetectorHit* RMGOutputTools::average_hits(std::vector<RMGDetectorHit*> hits,
...
+
+  RMGLog::Out(RMGLog::debug, "Merging gamma tracks ");
+
+
+  // for tracks of gammas look for a step close to each post-step point
+  // to redistribute the energy to
+  for (const auto& [trackid, input_hits] : hits_map) {
+
+    // only apply to gamma
+    if (input_hits.front()->particle_type != 22) continue;
```
gipert, line 124: I don't like to hardcode the particle id (22). I think there is a better way by using `G4Gamma`

gipert: like maybe https://geant4.kek.jp/Reference/11.3.0/classG4ParticleDefinition.html#a1d8cd4bb79a1161d62ec97f91ccf4a5a

src/RMGGermaniumOutputScheme.cc
```diff
@@ -296,6 +299,52 @@ void RMGGermaniumOutputScheme::DefineCommands() {
...
+          &RMGGermaniumOutputScheme::SetClusterDistance)
+      .SetGuidance("Set a distance threshold for the bulk pre-clustering.")
+      .SetParameterName("threshold", false)
+      .SetStates(G4State_Idle);
+  fMessenger
+      ->DeclareMethodWithUnit("SetPreClusterDistanceSurface", "um",
+          &RMGGermaniumOutputScheme::SetClusterDistanceSurface)
+      .SetGuidance("Set a distance threshold for the surface pre-clustering.")
+      .SetParameterName("threshold", false)
+      .SetStates(G4State_Idle);
```
gipert, line 325:
```suggestion
  fMessenger
      ->DeclareProperty("CombineLowEnergyElectronTracks", fPreClusterPars.combine_low_energy_tracks)
      .SetGuidance("Merge low energy electron tracks.")
      .SetStates(G4State_Idle);
  
  fMessenger->DeclareProperty("RedistributeGammaEnergy", fPreClusterPars.reassign_gamma_energy)
      .SetGuidance("Redistribute energy deposited by gamma tracks to nearby electron tracks.")
      .SetStates(G4State_Idle);

  fMessenger
      ->DeclareMethodWithUnit("SetPreClusterDistance", "um",
          &RMGGermaniumOutputScheme::SetClusterDistance)
      .SetGuidance("Set a distance threshold for the bulk pre-clustering.")
      .SetParameterName("threshold", false)
      .SetStates(G4State_Idle);
  
  fMessenger
      ->DeclareMethodWithUnit("SetPreClusterDistanceSurface", "um",
          &RMGGermaniumOutputScheme::SetClusterDistanceSurface)
      .SetGuidance("Set a distance threshold for the surface pre-clustering.")
      .SetParameterName("threshold", false)
      .SetStates(G4State_Idle);
```

src/RMGOutputTools.cc
```diff
@@ -0,0 +1,361 @@
...
+
+  // time from first hit
+  hit->global_time = hits.front()->global_time;
+
+  // The cluster represents a list of consecutive steps, so we can
+  // take the pre and post step from the first and last hit in the cluster.
+  hit->global_position_prestep = hits.front()->global_position_prestep;
+  hit->global_position_poststep = hits.back()->global_position_poststep;
+
+  // issue: this could be outside the volume!
```
gipert, line 93: this needs to be handled?

tdixon97: not sure, in practice its hard to see how it could happen?

tdixon97: do you have a solution ? I'm not sure about discarding the hit

gipert: also not sure how it can happen. maybe the statement is not true and we should just delete it

tdixon97: well it could eg. round the IC bore-hole corner?

gipert: ah I see you are right. what about aborting clustering in that case? does it make sense?

tdixon97: Thats an option, but a bit ugly

gipert: why ugly? it's the safest. will also happen very few times

tdixon97: Yeah I think I like your idea

src/RMGScintillatorOutputScheme.cc
```diff
@@ -15,6 +15,7 @@
 
 #include "RMGScintillatorOutputScheme.hh"
 
+#include <memory>
```
ManuelHu, line 18: add this to the header (actually, it is already needed there. We have `unique_ptr`s in the header, it was just forgotten)

also for germanium

include/RMGOutputTools.hh
```diff
@@ -0,0 +1,147 @@
...
+  };
+
+  /** @brief Get the position to save for a given hit.
+   *
+   * @details If the mode is @c kPostStep or if the particle is a gamma
+   * then the post-step point is used. This is since gamma's have only
+   * discrete interactions happening at the post-step.
+   * Otherwise if mode is @c kPreStep the prestep point is extracted, if it is either
+   * @c kBoth or @c kAverage the average of pre and post step is used (in the case of
+   * @c kBoth the pre and post step will also be saved separately).
```
ManuelHu, line 53:
```suggestion
   * @c kBoth the pre and post step needs to be saved separately).
```

src/RMGOutputTools.cc
ManuelHu: also above in line 34, and also with 11 for electrons below

ManuelHu (commented):
just some notes from a technical point of view

src/RMGScintillatorOutputScheme.cc
```diff
@@ -125,9 +154,13 @@ bool RMGScintillatorOutputScheme::ShouldDiscardEvent(const G4Event* event) {
...
 void RMGScintillatorOutputScheme::StoreEvent(const G4Event* event) {
   auto hit_coll = GetHitColl(event);
+
+  // pre-cluster the hits if requested
+  if (fPreClusterHits)
+    hit_coll = RMGOutputTools::pre_cluster_hits(hit_coll, fPreClusterPars, false, true);
+
   if (!hit_coll) return;
 
   if (hit_coll->entries() <= 0) {
```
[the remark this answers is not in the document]
tdixon97: Yeah I am not a fan of this deleting, I think I added it now, but its not a nice solution. Other solutions:
- something with std::shared_ptr,?
- we just copy the hit_coll into a `unique_ptr` (but then we have to allocate a bunch of new memory),
- we find some way to make `pre_cluster_hits` act in place on the hits, a bit hard since you loop over them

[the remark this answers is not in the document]
ManuelHu: the problem is that the original `hit_coll` cannot be deleted by us as it is managed by Geant4. So we cannot really put it in a smart pointer (note: It should technically be possible to create an unique_ptr without deleter that _would do nothing when going out of scope_, but that requires some ugly code.)

---
maybe something like this
```cpp
std::shared_ptr<HitCollection> _clustered_hits;
if (fPreClusterHits) {
   _clustered_hits = RMGOutputTools::pre_cluster_hits(...) // should now return a shred_ptr
   hit_coll = _clustered_hits.get(); // get an unmanaged ptr for use in this function
}
```
the shared_ptrshould ensure cleanup on exit without `delete` (also a bit ugly)

src/RMGOutputTools.cc
```diff
@@ -0,0 +1,351 @@
...
+      RMGLog::Out(RMGLog::debug, "Removing trackid ", trackid);
+    }
+  }
+  return output_hits;
+}
+
+RMGDetectorHitsCollection* RMGOutputTools::pre_cluster_hits(const RMGDetectorHitsCollection* hits,
+    RMGOutputTools::ClusterPars cluster_pars, bool has_distance_to_surface, bool has_velocity) {
+
+  // organise hits into a map based on trackid
```
ManuelHu, line 231: do we need a short-cut when `hits` is empty here? this is not checked in the output schemes before calling here.

tdixon97:
Added docs, what is still needed is to resolve the comments,  to decide on default options (and adjust all tests), and to handle the case the average point is outside the volume.

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+shape and they also have a different response close to the detector electrodes. So
+when simulating HPGe's it is advisable to save the information of the steps of
+particles within the detector. Then "post-processing" software such as _reboost_
+([docs](https://reboost.readthedocs.io/en/stable/)) can apply the detector response
+model without repeating the computationally intensive simulation.
+
+By default this output scheme writes out all steps in the registered sensitive
+HPGe detectors. The following properties of each hit are recorded (by default):
+
+- _time_: The global time of the hit,
```
gipert, line 107: I prefer monospaced font here
```suggestion
- `time`: The global time of the hit,
```

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+- _xloc_, _yloc_, _zloc_: the global position,
+- _evtid_ : the index of the Geant4 event,
+- _edep_: the deposited energy,
+- _dist_to_surf_: the distance of the hit from the detector surface.
+
+By default all floating point fields are saved with 64-bit (double) precision.
+The precision of the energy and or position / distance fields can be reduced
+to 32-bit with the macro commands:
+
+`/RMG/Output/Germanium/StoreSinglePrecisionPosition true`
```
gipert, line 118: Would be nice to directly link to the command reference every time (see #334)

gipert: Also, `true` is optional and not really making the command more clear here?

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+
+`/RMG/Output/Germanium/StoreSinglePrecisionEnergy true`
+
+It is possible to also store the track id and parent track id of each step with:
+
+`/RMG/Output/Germanium/StoreTrackID true`
+
+As mentioned earlier output schemes also provide a mechanism for filtering
+events. One useful option is to only write out events in which energy
+was deposited in a Germanium detector. This is used since the
```
gipert, line 130: Is Germanium capital because it's the name of the sensitive detector type? Then maybe use monospaced or italic font?

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+The precision of the energy and or position / distance fields can be reduced
+to 32-bit with the macro commands:
+
+`/RMG/Output/Germanium/StoreSinglePrecisionPosition true`
+
+and,
+
+`/RMG/Output/Germanium/StoreSinglePrecisionEnergy true`
+
+It is possible to also store the track id and parent track id of each step with:
```
gipert, line 124: what is the "track identifier"?

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+
+`/RMG/Output/Germanium/StoreTrackID true`
+
+As mentioned earlier output schemes also provide a mechanism for filtering
+events. One useful option is to only write out events in which energy
+was deposited in a Germanium detector. This is used since the
+other detector systems (LAr, water Cherenkov etc.) often act a "vetos's",
+so we are not interested in the energy deposited or steps if an event in the
+Germanium did not occur.
```
gipert, line 133:
```suggestion
germanium did not occur.
```

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+`/RMG/Output/Germanium/StoreSinglePrecisionEnergy true`
+
+It is possible to also store the track id and parent track id of each step with:
+
+`/RMG/Output/Germanium/StoreTrackID true`
+
+As mentioned earlier output schemes also provide a mechanism for filtering
+events. One useful option is to only write out events in which energy
+was deposited in a Germanium detector. This is used since the
+other detector systems (LAr, water Cherenkov etc.) often act a "vetos's",
```
gipert, line 131:
```suggestion
other detector systems (liquid argon, water Cherenkov etc.) often act a "vetos's",
```

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+so we are not interested in the energy deposited or steps if an event in the
+Germanium did not occur.
+The macro commands:
+
+`/RMG/Output/Germanium/AddDetectorForEdepThreshold {DET_UID}`
```
gipert, line 136: this is the syntax for variable substitutions, I think it's confusing to use it here. just use DET_UID

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+
+Finally, it is possible to "pre-cluster" the steps, this is used to reduce the
+amount of data written out to disk by combining steps very close together.
+Since the surface region of a HPGe detector has different properties to the bulk
+this clustering can be performed differently for surface and bulk hits
+(see [data-reduction](#data-reduction-methods) for more details).
+
+## Scintillator detectors
+
+This output scheme is used to record the steps in scintillation detectors (typically LAr),
```
gipert, line 181:
```suggestion
This output scheme is used to record the steps in scintillation detectors (typically liquid argon),
```

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+this is a calometric approach recording the energy deposited and steps. While the Optical
+output scheme is instead used for recording the detected optical photons.
+Most functionality is similar to the Germanium output scheme with a few exceptions.
+
+- Unlike for Germanium detectors the distance to the detector surface is not calculated,
+- The stacking possibility for optical tracks is not implemented,
+- The velocity of the particles can be saved using the
+
+`/RMG/Output/Scintillator/StoreParticleVelocities`
```
gipert, line 190: here and in all the rest: put everything on the same line and link to the macro command reference

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+
+macro command.
+
+## Data reduction methods
+
+Often Geant4 takes steps much shorter than those that are meaningful in a HPGe or a
+scintillation detector. For example the typical dimension of charge clouds
+produced by interactions in Germanium are 1-2 mm, so we are not sensitive to tracking at um levels.
```
gipert, line 198:
```suggestion
produced by interactions in Germanium are 1-2 mm, so we are not sensitive to tracking at the micrometer level.
```

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+To reduce the file size while retaining the useful information for computing observables of interest
+we have implemented some "pre-clustering" routines. These routines combine together steps that are very
+very close together.
```
gipert, line 201:
```suggestion
we have implemented some "pre-clustering" routines. These routines combine together steps that are very
very very very very very very very very very close together.
```

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+:::
+
+In order to have an efficient algorithm for pre-clustering we take use a "within-track" approach,
+this clusters only steps in the same `G4Track`, with some exceptions for very low energy tracks.
+In this way we only have to iterate through the steps in each event once. This also means the rows in our output are still interpretable
+with steps in the detector (just with a larger step length). The clustering is handled by the
+function:
+
+```C++
+  RMGDetectorHitsCollection* pre_cluster_hits(const RMGDetectorHitsCollection* hits,
```
gipert, line 215: also here link to the c++ api, see #334

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+      ClusterPars cluster_pars, bool has_distance_to_surface, bool has_velocity);
+```
+
+This takes in the pointer to the original `RMGDetectorHitsCollection` returning a pointer
```
gipert, line 219: also here missing API reference

docs/manual/output.md
```diff
@@ -91,3 +88,212 @@ expressed in the typical physical unit symbols. Unfortunately, column names
...
+:::{note}
+This design makes it easy to include additional clustering algorithms, a similar function just needs to be written.
+:::
+Pre-clustering can be enabled with the macro command:
+
+`/RMG/Output/Germanium/PreClusterOutputs True`
+
+and similar for the `Scintillator` output scheme.
+
+_remage_ first organises the hits by track id. Some processes in _Geant4_ produce
```
gipert, line 231:
```suggestion
_remage_ first organises the hits by track id. Some processes in Geant4 produce
```

gipert (changes requested):

src/RMGGermaniumOutputScheme.cc
```diff
@@ -296,6 +303,53 @@ void RMGGermaniumOutputScheme::DefineCommands() {
...
+      ->DeclareProperty("CombineLowEnergyElectronTracks", fPreClusterPars.combine_low_energy_tracks)
+      .SetGuidance("Merge low energy electron tracks.")
+      .SetStates(G4State_Idle);
+
+  fMessenger->DeclareProperty("RedistributeGammaEnergy", fPreClusterPars.reassign_gamma_energy)
+      .SetGuidance("Redistribute energy deposited by gamma tracks to nearby electron tracks.")
+      .SetStates(G4State_Idle);
+
+  fMessenger
+      ->DeclareMethodWithUnit("SetPreClusterDistance", "um",
```
gipert, line 320: I think we don't conventionally prepend `Set` to macro commands. Can you check?

docs/manual/output.md
tdixon97: track id

gipert: i meant maybe two words about where this id comes from

src/RMGScintillatorOutputScheme.cc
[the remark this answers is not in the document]
tdixon97: do you think its preferable to just deleting? Maybe a bit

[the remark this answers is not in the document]
ManuelHu: At least we can avoid forgetting to delete it, or deleting it twice...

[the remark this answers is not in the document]
ManuelHu: maybe it is shorter to wrap the original hit_coll in a shared_ptr with an empty deleter, and just swith to shared_ptr usage in the whole function: https://stackoverflow.com/questions/20131877/how-do-you-make-stdshared-ptr-not-call-delete

src/RMGOutputTools.cc
```diff
@@ -0,0 +1,399 @@
...
+
+        cluster_first_hit = hit;
+      } else {
+        hits_vector.back().push_back(hit);
+      }
+    }
+  }
+
+  // create a container for the output hits
+  auto out = std::shared_ptr<RMGDetectorHitsCollection>();
```
ManuelHu, line 350:
```suggestion
  auto out = std::make_shared<RMGDetectorHitsCollection>();
```

with the other you constructed a nullptr, I guess?

gipert (approved):
there's still a couple of minor things i'd like to change in the docs but i can do that in the future

tdixon97:
Why dont you change them quickly now? Later we will forget

docs/manual/output.md
gipert, line 1: @tdixon97 have a look

PR #332, opened 2025-04-18
Add jupyter install guide

EricMEsch:
First draft for a jupyter install guide.

Missing:

- [ ] Docker installation guide. (I tried but did not get anything working so far. I honestly do not even understand what the issue with docker is)

### DIFF
--- a/docs/manual/install.md
+++ b/docs/manual/install.md
@@ -50,3 +50,273 @@ $ mkdir build && cd build
 $ cmake -DCMAKE_INSTALL_PREFIX=<unique prefix> ..
 $ make install
 ```
+
+## Setting up Jupyter
+
+Since the latest remage versions, it is possible to include remage into your jupyter python kernel. This allows you to run everything within a single jupyter notebook. Because the setup is very easy, this might be more comfortable and easier than using the CLI (depending on your experience and use-case). Keep in mind that interactive visualization, like the Pygeometry or Geant4 visualizer, might be very buggy in a notebook. You can still try it, but you can not say that we did not warn you!
+
+To set up your jupyter kernel, you only have to provide the correct `kernel.json` file to your jupyter installation. This means you first have to find your jupyters share folder and create the `kernel.json` file there. 
+```console
+cd ~/.local/share/jupyter
+mkdir kernels
+cd kernels
+mkdir remage
+cd remage
+touch kernel.json
+```
+
+The contents of this kernel file now depends on how you installed remage.
+
+:::{note} If you are using a jupyter that is installed inside of a virtual environment (or in vscode your default python interpreter is inside a venv) then you probably need to add the `kernel.json` to the share folder of that jupyter installation instead.
+:::
+
+### Jupyter from pre-built binaries
+
+If you have installed remage from a pre-built binary using apptainer, your kernel file should look something like this (remember to replace `/path/to/remage_latest.sif` with your actual path):
+
+```json
+{
+ "argv": [
+  "apptainer",
+  "exec",
+  "-B", "/run/user/1000:/run/user/1000",
+  "/path/to/remage_latest.sif",
+  "python",
+  "-m",
+  "ipykernel_launcher",
+  "-f",
+  "{connection_file}"
+ ],
+ "display_name": "remage-latest (apptainer)",
+ "language": "python",
+ "metadata": {
+  "debugger": true
+ }
+}
+```
+
+If you instead installed using docker it should look something like this: I did not get this working yet :cry:
+
+### Jupyter from source
+
+We are assuming that you have built remage using the `remage-base_latest` container. In this case, after `make install`-ing remage, the kernel file is almost the same as if you used the pre-built library. You only have to additionally tell the kernel where you installed remage with the export paths. The main issue here is, that `env` does not expand shell variables like `$PATH`, it only overwrites them. So we have to set them up ourself. The first path for "PATH=..." and "LD_LIBRARY_PATH..." should be the path where you installed remage. (Replace `remage/install/path/...` with the path the terminal told you to export when you did `cmake ..`). Also remember to replace `/path/to/remage_base_latest.sif` with the actual path of your image.
+
+```json
+{
+  "argv": [
+    "apptainer",
+    "exec",
+    "-B", "/run/user/1000:/run/user/1000",
+    "/path/to/remage_base_latest.sif",
+    "env",
+    "PATH=/remage/install/path/bin:/opt/python-extra/bin:/opt/bxdecay0/bin:/opt/geant4/bin:/opt/root/bin:/opt/vecgeom/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+    "LD_LIBRARY_PATH=/remage/install/path/lib:/opt/bxdecay0/lib:/opt/geant4/lib:/opt/root/lib:/opt/vecgeom/lib::/.singularity.d/libs",
+    "/opt/python-extra/bin/python",
+    "-m", "ipykernel_launcher",
+    "-f", "{connection_file}"
+  ],
+  "display_name": "remage-base (apptainer)",
+  "language": "python",
+  "metadata": {
+    "debugger": true
+  }
+}
+```
+
+Note that this will overwrite the shell environment variables `$PATH` and `$LD_LIBRARY_PATH`. 
+
+Because that is not the prettiest solution, you could also write a small setup script instead.
+
+:::{note} It already works with the first solution. This is an alternative solution now.
+:::
+
+For this you should write a small bash script in a folder of your liking.
+
+```console
+cd # To go back to your home directory
+mkdir scripts
+cd scripts
+touch remage_kernel_setup.sh
+chmod +x remage_kernel_setup.sh
+```
+
+In the setup script `remage_kernel_setup.sh` you should now write something like this:
+```bash
+#!/bin/bash
+export PATH=/remage/install/path/bin:$PATH
+export LD_LIBRARY_PATH=/remage/install/path/lib:$LD_LIBRARY_PATH
+
+exec /opt/python-extra/bin/python -m ipykernel_launcher -f "$1"
+```
+Where again `/remage/install/path/` should be replaced with the path where you installed remage.
+Your `kernel.json` should then be adjusted to look like this:
+```json
+{
+  "argv": [
+   "apptainer",
+   "exec",
+   "-B", "/run/user/1000:/run/user/1000",
+   "-B", "/home/user/scripts/remage_kernel_setup.sh:/opt/remage/remage_kernel_setup.sh",
+   "/path/to/remage_base_latest.sif",
+   "/opt/remage/remage_kernel_wrapper.sh",
+   "{connection_file}"
+  ],
+  "display_name": "remage-base (apptainer)",
+  "language": "python",
+  "metadata": {
+   "debugger": true
+  }
+}
+```
+Remember to replace `/path/to/remage_base_latest.sif` with the actual path of your image and `/home/user/scripts/remage_kernel_setup.sh` with the actual path of the script!
+### Jupyter on NERSC
+
+Setting up the remage jupyter kernel on NERSC is exactly the same as setting up the legend-sw kernel on NERSC like explained [on Confluence](https://legend-exp.atlassian.net/wiki/spaces/LEGEND/pages/261750878/Computing+at+NERSC#Python-notebooks-in-Shifter-containers). In order to make sure the kernel does not have to first pull the remage version from Docker, maybe lets first pull it ourselves:
+```console
+shifterimg -v pull legendexp/remage:latest
+```
+The rest is literaly the same as in the Confluence explanation:
+```console
+$ cd ~/.local/share/jupyter
+$ mkdir kernels
+$ cd kernels
+$ mkdir legend-base
+$ cd legend-base
+$ touch kernel.json
+```
+The only difference to the Confluence version is that in the `kernel.json` instead of "--image=legendexp/legend-software:latest" we write "--image=legendexp/remage:latest" and instead of "/opt/conda/bin/python3" we write "python". Now maybe also change the name from ""display_name": "legend-sw"," to ""display_name": "remage",". In the end your `kernel.json` should look like this:
+```json
+  {
+  "argv": [
+    "shifter",
+    "--module=none",
+    "--image=legendexp/remage:latest",
+    "python",
+    "-m",
+    "ipykernel_launcher",
+    "-f",
+    "{connection_file}"
+  ],
+  "display_name": "remage",
+  "language": "python",
+  "env": {
+    "LEGENDDATADIR": "/global/cfs/projectdirs/m2676/data"
+  }
+}
+```
+Now after refreshing JupyterLab you should see the "remage" kernel in your notebooks!
+
+### Example Jupyter file for testing
+
+To test if it works, open up a new notebook and select the remage kernel! You might have to search for it under "Select another Kernel" "Jupyter kernel..." depending on your installation, the name should be equal to the name under "display_name" in the `kernel.json`. Now first start with a simple cell to check if the kernel starts succesfully:
+```python
+# Change this to the path where you want to save simulation geometry and data
+path = "."
+```
+Make this a new code cell and run it. If everything worked the kernel should load up fine (might take some more time the first time.). Now lets see if the pygeometry dependencies will be found and we can build a geometry:
+
+```python
+from __future__ import annotations
+
+import legendhpges as hpges
+import pyg4ometry as pg4
+import numpy as np
+
+reg = pg4.geant4.Registry()
+
+# create a world volume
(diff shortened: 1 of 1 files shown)


EricMEsch:
addresses part of #324

tdixon97:
This goes a bit beyond what we might normally put in an "installation guide" - what about making it a seperate page?

tdixon97:
Or even having the basic (minimal?) info in the user manual / installation guide and having an example in a page of tutorials / examples? I would consider this really an advanced option

EricMEsch:
Well i dont mind where to place this. Having a seperate page might be better

tdixon97:
Yep, lets see what @gipert and @ManuelHu think, but overall it looks to me very good.
What about the minimum information (not the examples) in the installation page and a tutorial / example elsewhere?

docs/manual/install.md
```diff
@@ -50,3 +50,312 @@ $ mkdir build && cd build
...
+The contents of this kernel file now depends on how you installed remage.
+
+:::{note} If you are using a jupyter that is installed inside of a virtual environment (or in vscode your default python interpreter is inside a venv) then you probably need to add the `kernel.json` to the share folder of that jupyter installation instead.
+:::
+
+### Jupyter from pre-built binaries
+
+If you have installed remage from a pre-built binary using apptainer, your kernel file should look something like this (remember to replace `/path/to/remage_latest.sif` with your actual path):
+
+```json
```
tdixon97, line 78: What about having them as a JSON file hidden unless you click on it (to avoid clutter), see the macro file in the tutorial for an example

docs/manual/install.md
```diff
@@ -50,3 +50,312 @@ $ mkdir build && cd build
...
+}
+```
+
+If you instead installed using docker it should look something like this: I did not get this working yet :cry:
+
+### Jupyter from source
+
+We are assuming that you have built remage using the `remage-base_latest` container and followed the explanation [above](#building-from-source). In this case, after `make install`-ing remage, the kernel file is almost the same as if you used the pre-built library. You only have to additionally tell the kernel where you installed remage with the export paths. The main issue here is, that `env` does not expand shell variables like `$PATH`, it only overwrites them. So we have to set them up ourself. The first path for `"PATH=..."` and `"LD_LIBRARY_PATH..."` should be the path where you installed remage. (Replace `remage/install/path/...` with the path the terminal told you to export when you did `cmake ..` during the installation). Also remember to replace `/path/to/remage_base_latest.sif` with the actual path of your image.
+
+```json
```
tdixon97, line 106: similar to above?

docs/manual/install.md
```diff
@@ -50,3 +50,312 @@ $ mkdir build && cd build
...
+  "language": "python",
+  "metadata": {
+    "debugger": true
+  }
+}
+```
+
+Remember to replace `/path/to/remage_base_latest.sif` with the actual path of your image and `/home/user/scripts/remage_kernel_setup.sh` with the actual path of the script!
+
+### Jupyter on NERSC
```
tdixon97, line 184: I wonder if this belong in the remage documentation?

EricMEsch: Well it is still about running remage in a notebook on NERSC

docs/manual/install.md
```diff
@@ -50,3 +50,312 @@ $ mkdir build && cd build
...
+  "language": "python",
+  "env": {
+    "LEGENDDATADIR": "/global/cfs/projectdirs/m2676/data"
+  }
+}
+```
+
+Now after refreshing JupyterLab you should see the "remage" kernel in your notebooks!
+
+### Example Jupyter file for testing
```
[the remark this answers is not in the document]
EricMEsch: Is it possible to use remage_run now from an external environment?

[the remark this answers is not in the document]
ManuelHu: no

docs/manual/install.md
tdixon97: I think that is super LEGEND specific and maybe belongs on confluence instead

docs/manual/install.md
```diff
@@ -50,3 +50,312 @@ $ mkdir build && cd build
...
+
+```console
+$ cd # To go back to your home directory
+$ mkdir scripts
+$ cd scripts
+$ touch remage_kernel_setup.sh
+$ chmod +x remage_kernel_setup.sh
+```
+
+In the setup script `remage_kernel_setup.sh` you should now write something like this:
```
ManuelHu, line 148: we could also ship this file within the container?

tdixon97: I think we should either provide the file or give more basic instructions? I seems a bit too much to me to tell people this.

ManuelHu: I had this thought at the beginning, too. But we inherently cannot provide this file - this is just for people compiling remage from source. SO the path is not known at the beginning

docs/manual/install.md
```diff
@@ -50,3 +50,312 @@ $ mkdir build && cd build
 $ cmake -DCMAKE_INSTALL_PREFIX=<unique prefix> ..
 $ make install
 ```
+
+## Setting up Jupyter
+
+Since the latest remage versions, it is possible to include remage into your jupyter python kernel. This allows you to run everything within a single jupyter notebook. Because the setup is very easy, this might be more comfortable and easier than using the CLI (depending on your experience and use-case). Keep in mind that interactive visualization, like the Pygeometry or Geant4 visualizer, might be very buggy in a notebook. You can still try it, but you can not say that we did not warn you!
```
tdixon97, line 56: pyg4ometry (correct spelling and case) works pretty well in the notebook. I do not know how well Geant4 in interactive mode will work

docs/manual/install.md
ManuelHu: Maybe this should really be split into a "simple installation" description using the pre-built container images; and an "install from source/advanced" file to not confuse people?

tdixon97: Maybe a good idea

docs/manual/install.md
gipert: I would not hide it

docs/manual/install.md
```diff
@@ -50,3 +50,312 @@ $ mkdir build && cd build
...
+
+If you have installed remage from a pre-built binary using apptainer, your kernel file should look something like this (remember to replace `/path/to/remage_latest.sif` with your actual path):
+
+```json
+{
+  "argv": [
+    "apptainer",
+    "exec",
+    "-B",
+    "/run/user/1000:/run/user/1000",
```
gipert, line 84: This should be replaced by $XDG_RUNTIME_DIR:
```suggestion
    "$XDG_RUNTIME_DIR:$XDG_RUNTIME_DIR",
```
but you should check whether it works or not. Actually, is this even needed?

docs/manual/install.md
```diff
@@ -50,3 +50,312 @@ $ mkdir build && cd build
...
+    "/path/to/remage_latest.sif",
+    "python",
+    "-m",
+    "ipykernel_launcher",
+    "-f",
+    "{connection_file}"
+  ],
+  "display_name": "remage-latest (apptainer)",
```
gipert, line 92:
```suggestion
  "display_name": "remage container",
```

docs/manual/install.md
```diff
@@ -50,3 +50,312 @@ $ mkdir build && cd build
...
+Since the latest remage versions, it is possible to include remage into your jupyter python kernel. This allows you to run everything within a single jupyter notebook. Because the setup is very easy, this might be more comfortable and easier than using the CLI (depending on your experience and use-case). Keep in mind that interactive visualization, like the Pygeometry or Geant4 visualizer, might be very buggy in a notebook. You can still try it, but you can not say that we did not warn you!
+
+To set up your jupyter kernel, you only have to provide the correct `kernel.json` file to your jupyter installation. This means you first have to find your jupyters share folder and create the `kernel.json` file there.
+
+```console
+$ cd ~/.local/share/jupyter
+$ mkdir kernels
+$ cd kernels
+$ mkdir remage
+$ cd remage
```
gipert, line 65:
```suggestion
$ mkdir -p ~/.local/share/jupyter/kernels/remage
$ touch ~/.local/share/jupyter/kernels/remage/kernel.json
```

docs/manual/install.md
EricMEsch: But it also is super easy and i would like people to find it... As they literally only have to replace a few lines of the stuff that they probably already have...

I am not sure how active people search for stuff on confluence, but maybe you are right

docs/manual/install.md
EricMEsch: But in that case we might also split the installation guide at that point?
Like have one page with the installation for pre-built container & jupyter setup
And then a more advanced page with the installation from source & jupyter setup

docs/manual/install.md
EricMEsch: Does the interactive thing work for you without being buggy? For me it works, but i can not close it. Like i have to close vscode in order for the window to go away

docs/manual/install.md
EricMEsch: For me it was. Without this it did not work

docs/manual/install.md
tdixon97: What do you think @gipert, I see your point Eric but I think if we start explaining the installation on every system it could get a bit too much. Eg. now there is also no explanation of the "standard" shifter installation for NERSC

docs/manual/install.md
EricMEsch: I just used the way it was shown on Confluence, :D

docs/manual/install.md
gipert: I think this jupyter part is too long, I would really just show the basics (the kernel file) and pointers to existing external documentation. And the example of how to run remage

docs/manual/install.md
ManuelHu: If this doies not expand shell variables, you really have to document where to get this. The user id might not be `1000` on all linux distributions, not even speaking about shared compute servers.

This line would probably fail when I run it on our compute server's jupyter instance with UID != 1000.

ManuelHu: just for reference: Oliver's cenv also bind-mounts this directory, so this really might really be necessary? https://github.com/oschulz/container-env/blob/0c9b80016dcf6b3ba922de38689e849815ae8430/bin/cenv#L401-L403

ManuelHu: more:

https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs

states that `${VAR_NAME}` can be used to be substituted, and not `$VAR_NAME`. This could simplify all this (also the $PATH/$LD_LIBRARY_PATH below?)
please test

EricMEsch: Oh thanks, that sounds great, i will test next week!

EricMEsch:
I have reduced it to a short and basic version now

EricMEsch:
This would be the option to add a jupyter example to the manual... Not sure if we want it where i placed it

docs/manual/jupyter.md
```diff
@@ -0,0 +1,135 @@
...
+hist.new.Reg(440, 0, 2200, name="energy [keV]").Double().fill(
+    ak.sum(evt.edep, axis=-1)
+).plot(yerr=False)
+plt.ylabel("counts / 5 keV")
+plt.xlim(500, 2200)
+plt.yscale("log")
+```
+
+If everything worked you should now be able to run the full simulation chain in a notebook, like you did just now!
+![Happy visualization](https://media.tenor.com/lCKwsD2OW1kAAAAj/happy-cat-happy-happy-cat.gif)
```
ManuelHu, line 135: Not sure if we want to have gifs in the documentation...?

EricMEsch: image

tdixon97:
I think this is all very nice, one thing to think about for future. If we want to go with this idea of running remage directly from python then we could make a nicer API, I thought that some aspects of Geant4.jl are nice (eg. [link](https://juliaphysics.github.io/SolidStateDetectors.jl/main/tutorials/geant4_ssd/))

ManuelHu:
we will not have a way to pass structured data from python to the cpp side, except for cli args

tdixon97:
Sure but you could imagine some other `remage_run` functions that internally just product the correct `remage` arguments (basically the macro). I think it might be a nice feature for simple things. Another feature I think that might be very nice would be returning the `lh5` object in memory. How hard would that be?

EricMEsch:
So in theory the docker version works now.

1. For some reason, even when you set the `--rm` flag the docker container will not terminate if you do not manually terminate it. This means everytime you restart the kernel you get another container running which will quickly lead to something like this:
image Does anyone have experience with this? I have tried a bunch of options within the `kernel.json` but i do not think they change anything at all. I could try a setup script again...

2. I do not have access to a mac, so i am not sure if this would work at all on mac. I kind off doubt it, because the `--network=host` option seemed to be crucial. 

Also do we care about this docker version? I guess the only reason would be to have it supported for macOS...

EricMEsch:
The main questions are

- Do we care about the docker version? (as the way it currently works is highly questionable)
- Do we care about a solution for macOS?
- Do we want the jupyter example somewhere?

tdixon97:
- If docker isn't working well can we not just tell users its not supported?
- For mac for now I think we just say its not supported,
The example I think is useful, I think what we can do in future is have a page of examples/ advanced tutorials? What do you think @gipert ?

tdixon97:
Btw, I think this is still a bit too long, the installation manual shouldn't include code its self, just the basic info

ManuelHu:
In the default venv inside the container, the `remage` package is not installed

tdixon97:
unless we have the remage package I don't find the concept of running sims in Jupyter notebook useful (i.e. if it involves calling subprocess what was the gain..)
Can we change that @ManuelHu ?

ManuelHu:
> Can we change that @ManuelHu ?

In the last commit I added this to the dockerfile, lets wait and see if the dev container will build (and work)

docs/manual/install.md
```diff
@@ -50,3 +50,85 @@ $ mkdir build && cd build
...
+  "display_name": "remage (docker-dev)",
+  "language": "python",
+  "metadata": {
+    "debugger": true
+  }
+}
+```
+
+:::{note} On macOS `--network=host` is not supported. Try `"docker run --rm --entrypoint python -v $(pwd):/work -v $HOME/.jupyter/runtime:/root/.jupyter/runtime -w /work legendexp/remage:dev -m ipykernel_launcher -f {connection_file}"` instead. This is untested and not sure if it works.
+:::
```
gipert, line 103: we cannot test this so I would just remove it. we can add a note in the future if anyone gives it a try on macos

docs/manual/install.md
```diff
@@ -50,3 +50,85 @@ $ mkdir build && cd build
...
+
+:::{warning} Upon start, the kernel will start a new docker container, but for some reason the container will never be terminated, unless you manually os.\_exit(0) from within the kernel.
```
gipert, line 105: I will give this a try

docs/manual/install.md
```diff
@@ -50,3 +50,85 @@ $ mkdir build && cd build
...
+:::
+
+Now after refreshing JupyterLab/VSCode you should be able to find the "remage" kernel in your notebooks! You might have to search for it under "Select another Kernel" "Jupyter kernel..." depending on your installation, the name should be equal to the name under `display_name` in the `kernel.json`.
```
gipert, line 108:
```suggestion
Now after refreshing JupyterLab/VSCode you should be able to find the "remage" kernel in your notebooks!
```
not needed, let's try to keep this part of the docs short

docs/manual/install.md
```diff
@@ -50,3 +50,85 @@ $ mkdir build && cd build
...
+
+Now let's check if everything works, you should have a working `geometry.gdml` and `gammas.mac` file from the basic tutorial:
```
gipert, line 110: we don't need and example here, i would remove all the following part

docs/manual/jupyter.md
gipert, line 1: I would also remove this, as tutorials are not specific to jupyter

docs/manual/install.md
```diff
@@ -50,3 +50,36 @@ $ mkdir build && cd build
...
+  ],
+  "display_name": "remage container",
+  "language": "python",
+  "metadata": {
+    "debugger": true
+  }
+}
+```
+
+Now after refreshing JupyterLab/VSCode you should be able to find the "remage" kernel in your notebooks!
```
ManuelHu, line 85:
```suggestion
Now after refreshing JupyterLab/VSCode you should be able to find the "remage container" kernel in your notebooks!
```

this should be the display_name, right?

docs/manual/install.md
```diff
@@ -84,4 +89,10 @@ If you have installed remage from a pre-built binary using apptainer, your kerne
 }
 ```
 
-Now after refreshing JupyterLab/VSCode you should be able to find the "remage container" kernel in your notebooks!
+Now after refreshing JupyterLab/VSCode you should be able to find the "remage
+container" kernel in your notebooks!
+
+:::{warning}
+Interactive visualization, like the pyg4ometry or Geant4 visualizer, might be
+not work as expected in a notebook.
```
EricMEsch, line 97: "might be not work"

Issue #317, opened 2025-04-02
rework remage-cpp discovery


ManuelHu:
Currently the remage python package contains hardcoded paths. This is okay or even necessary in some contexts:
* testing of non-installed remage binaries using ctest (necessary)
* installed packages in the container image

I could imaging having another, _addition_ mechanism to find `remage-cpp` from python: have a default way to discover it in `$PATH` when no explicit path is hardcoded into the installed package.
This package could then be uploaded to pypi and easily installed by users into their venvs.

this would also require having a strict version check that the python-wrapper can only be used with the corresponding C++ binaries.

tdixon97:
I am not sure about the details but long term this seems a good idea. Maybe not a high priority now though

ManuelHu:
note: I already have half of the needed code (extremely untested and probably not working)

PR #337, opened 2025-04-22
python: add find_remage module

ManuelHu:
This should not change functionality when installed via cmake, but should allow to install the python package directly from the source tree, without using cmake.

This PR now makes the wrapper actually installable via pip/pypi (`pip install .` in the repository root, not in the python sub-directory). sdists and wheels are adjusted to be buildable in both ways and only contain the python code.

**note:** when trying this out, make sure that no `cpp_config.py` file exists in the folder when installing, as that would disable the new path-based discovery logic. This file should not be persistent in any case, after you ran `make` the first time.

see #317

### DIFF
--- a/.github/workflows/distribution.yml
+++ b/.github/workflows/distribution.yml
@@ -1,0 +1,30 @@
+name: distribute
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
+  release:
+    types:
+      - published
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  FORCE_COLOR: 3
+
+jobs:
+  dist:
+    name: Distribution build
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: hynek/build-and-inspect-python-package@v2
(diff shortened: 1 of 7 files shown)


gipert:
very nice. seems promising that checks are passing?

ManuelHu:
this is the main part of the new logic:

https://github.com/ManuelHu/remage/blob/72e98473b62f34bc8adb41ff7ae0e02ab723cae4/python/remage/find_remage.py#L22-L31

ManuelHu:
this should now work better for manual testing (i.e. it is possible to do a `pip install .` in the repository root to test a python package with path-based discovery.
It is just important that the python/venv bin-directory has a high priority in $PATH then the remage install dir. Both have to be in path, however.

ManuelHu:
this will still remain a draft for a while, I guess. The version checks works but is unstable - i.e. the python version and cpp version might diverge after doing local development, leading to built binaries not working (this cannot happen for builds from a clean source checkout, so CI is fine)

ManuelHu:
This should really work now. it would be nice if you could test this in your environments @EricMEsch @tdixon97

1. create a new venv
2. install remage with `pip install .` into that env (execute in repo root)
3. setup the paths so that your new venv takes precendence over the remage (cpp+python) install location
4. try to run `remage`
5. try to run `REMAGE_ASSERT_CPP_ORIGIN=path remage`
6. try to use `remage_run` from an  interactive python shell/jupyter from the new env

ManuelHu:
I would merge this now. In the current state, this still requires some minor work to actually get it on pypi. For the existing cmake-based installation, it does not change anything

PR #334, opened 2025-04-21
Populate user manual

gipert:

PR #339, opened 2025-04-28
Allow surface sampling on volumes with daughters

tdixon97:
I think this should work ... events will just be located on the mother volume.


ManuelHu:
you mean on the outside of the mother volume? I.e. the wording is not so clear, the surface of the mother volme is also on the inside...

tdixon97:
We can easily if it worked from the validation suite when that finishes running

gipert:
yes, we need this feature. I have added a note in #334

PR #341, opened 2025-04-28
Merge threads and reshape

tdixon97:
Needs the update from reboost to be merged and some more info from the ipc from C++ (which @ManuelHu says is easy)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-  "legend-pydataobj==1.11.7",
+  "legend-pydataobj<1.12",
   "colorlog",
   "tol_colors",
   "reboost>=0.2.4",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
   "legend-pydataobj==1.11.7",
   "colorlog",
   "tol_colors",
-  "reboost>=0.2.3",
+  "reboost>=0.2.4",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-  "legend-pydataobj",
+  "legend-pydataobj==1.11.7",
   "colorlog",
   "tol_colors",
   "reboost>=0.2.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
   "legend-pydataobj==1.11.7",
   "colorlog",
   "tol_colors",
-  "reboost>=0.2.1",
+  "reboost>=0.2.3",
 ]
 
 [project.optional-dependencies]
--- a/python/remage/cli.py
+++ b/python/remage/cli.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import argparse
 import contextlib
 import logging
 import os
@@ -8,12 +9,12 @@ import signal
 import subprocess
 import sys
 import threading
-import argparse
 from collections.abc import Iterable
 from pathlib import Path
 
 import colorlog
 from lgdo import lh5
+from reboost.build_hit import build_hit
 
 from .find_remage import find_remage_cpp
 from .ipc import IpcResult, ipc_thread_fn
@@ -287,21 +288,35 @@ def remage_run_from_args(
         see :meth:`remage_run`
     """
     logger = _setup_log()
-    
+
     parser = argparse.ArgumentParser(allow_abbrev=False, add_help=False)
-    
+
     # python only arguments
-    parser.add_argument("-m","--merge-output-files", action="store_true",
-        help="Flag to merge output files")
-    parser.add_argument("-r","--reshape-output", action="store_true",
-        help="Flag to reshape output steps so the output is `hit` oriented.")
-    parser.add_argument("-T","--time-window-in-us", required=False,type=float,
-        help="Time window (in us) to group together steps for reshping")
-    
+    parser.add_argument(
+        "-m",
+        "--merge-output-files",
+        action="store_true",
+        help="Flag to merge output files",
+    )
+    parser.add_argument(
+        "-r",
+        "--reshape-output",
+        action="store_true",
+        help="Flag to reshape output steps so the output is `hit` oriented.",
+    )
+    parser.add_argument(
+        "-T",
+        "--time-window-in-us",
+        required=False,
+        type=float,
+        metavar="",
+        help="Time window in microseconds to group steps for reshaping.",
+    )
+
     py_args, cpp_args = parser.parse_known_args(args)
-    
+
     ec, termsig, ipc_info = _run_remage_cpp(cpp_args, is_cli=args is None)
-    
+
     # print an error message for the termination signal, similar to what bash does.
     if termsig not in (None, signal.SIGINT, signal.SIGPIPE):
         log.error(
@@ -348,8 +363,8 @@ def remage_run_from_args(
     # output post-processing (merging multiple LH5 files)
     output_files = ipc_info.get("output")
     main_output_file = ipc_info.get_single("output_main", None)
-    overwrite_output = ipc_info.get_single("overwrite_output", "0") == "1"  # noqa: F841
-    registered_detectors = ipc_info.get("detector", 3)  # noqa: F841
+    overwrite_output = ipc_info.get_single("overwrite_output", "0") == "1"
+    registered_detectors = ipc_info.get("detector", 3)
 
     # we might have no output file.
     if len(output_files) > 1 and main_output_file is not None:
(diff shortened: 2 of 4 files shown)


src/remage.cc
```diff
@@ -85,6 +89,11 @@ int main(int argc, char** argv) {
       ->type_name("FILE");
   app.add_option("-o,--output-file", output, "Output file for detector hits")->type_name("FILE");
   app.add_flag("-w,--overwrite", overwrite_output, "Overwrite existing output files");
+  app.add_flag("-m,--merge", merge_out_files, "Merge output files");
```
gipert, line 92: don't add wrapper args to the C++ program, add them to the wrapper via ArgParse and then use `parse_known_args()` to extract them and pass a correct argument list to remage-cpp

ManuelHu: @tdixon97 I already have working code for that when I tested this with Luigi. I will send you an example patch that should work

gipert:
yes

tdixon97:
for some reason the application now takes ~ 30s to start for me now. Not sure if its related to the change I made since I realise that the path for `remage` my system was using was not the one in the venv... (some issue?)

ManuelHu:
> one issue is then they dont get passed to remage_run directly

I do not really understand what you mean.

The changes in my path are in `_remage_run_from args()` which is the main logic taking an array of arguments and doing everything necessary (including post-processing). This is _below_ `remage_run_from_args()`, so the new arguments can automatically be used with that.

To use them with the kwargs-style API `remage_run()` they obviously also need to be defined there.

> maybe instead this extra python argparser should be in remage_cli ?

No. `remage_cli` is only called when you run remage _directly_ on CLI. You probably want to be able to pass your arguments from `remage_run[_from_args]()` at some point, no? `remage_cli()` should stay the same one-line wrapper as it currently is.

ManuelHu:
please also rebase on current main, that contains the mentioned C++ changes to send detector info over ipc

python/remage/cli.py
```diff
@@ -264,26 +361,43 @@ def remage_run_from_args(
...
+                #    config
+                #    {},
+                #    stp_files = output_files,
+                #    glm_files = None,
+                #    hit_files = main_output_file
+                # )
+                msg = "Reshaping output files is not implemented yet"
+                raise NotImplementedError(msg)
+
+            lh5.concat.lh5concat(
```
gipert, line 387: the logic of this block seems wrong, please review. also info message about merging in progress?

python/remage/cli.py
```diff
@@ -181,6 +239,15 @@ def remage_run(
...
     args.append(f"--threads={threads}")
 
+    if merge_output_files:
+        args.append("--merge-output-files")
+
+    if reshape_output_files:
+        args.append("--reshape-output")
+
+    if time_window is not None:
+        args.append(f"--time-window-in-us {time_window}")
```
ManuelHu, line 249:
```suggestion
        args.append(f"--time-window-in-us={time_window}")
```

otherwise it won't work - `args` is already an array of space-delimited strings.

src/remage.cc
ManuelHu, line 1: please revert unrelated changes to this file

python/remage/cli.py
```diff
@@ -12,13 +13,61 @@
...
+from lgdo import lh5
+from reboost.build_hit import build_hit
 
 from .find_remage import find_remage_cpp
 from .ipc import IpcResult, ipc_thread_fn
 
 log = logging.getLogger(__name__)
 
 
+def get_rebooost_config(
```
ManuelHu, line 25: at some point we should move the post-processing logic to a seperate file to make the code easier to understand, but for now it should be fine.

ManuelHu:
> need to get the tables that are not detectors (eg vertices) from c++

but this should be just the difference of tables in the file and sensitive detectors?

python/remage/cli.py
gipert: yes this is important. can you make a post processing module @tdixon97?

python/remage/cli.py
```diff
@@ -234,6 +328,11 @@ def remage_run_from_args(
     # clean-up should run always, irrespective of exit code.
     _cleanup_tmp_files(ipc_info)
 
+    if "-h" in cpp_args or "--help" in cpp_args:
+        print()  # noqa: T201
+        print("PYTHON-ONLY OPTIONS:")  # noqa: T201
```
gipert, line 333: I wonder if we want to name this section. in the end how the remage executable works is a detail for the user.

tdixon97: I agree

tdixon97:
any ideas why the mac tests fail?

gipert:
cool error. doesn't seem our fault

pyproject.toml
```diff
@@ -32,7 +32,7 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-  "legend-pydataobj",
+  "legend-pydataobj==1.11.7",
```
gipert, line 35:
```suggestion
  "legend-pydataobj>=1.11.7",
```

python/remage/cli.py
```diff
@@ -155,6 +163,12 @@ def remage_run(
         set the number of threads used by remage
     overwrite_output
         overwrite existing output files
+    merge_output_files
+        merge output files from individual threads.
+    reshape_output_files
+        reshape output files to be hit oriented.
```
gipert, line 169: hard to understand what this description means

python/remage/cli.py
```diff
@@ -262,29 +314,84 @@ def remage_run_from_args(
...
+                    config,
+                    {},
+                    stp_files=original_files,
+                    glm_files=None,
+                    hit_files=output_files,
+                )
+
+            # else use lh5concat
+            elif py_args.merge_output_files:
+                msg = "Begin merging output files"
```
gipert, line 371:
```suggestion
                msg = "Merging output files"
```

python/remage/cli.py
```diff
@@ -262,29 +314,84 @@ def remage_run_from_args(
...
-            #     overwrite=overwrite_output,
-            # )
-            # set the merged output file for downstream consumers.
-            # ipc_info.set("output", [main_output_file])
-            # delete un-merged output files.
-            # for f in output_files:
-            #     Path(f).unlink()
-    elif len(output_files) > 1:
+            msg = "Begin python based post-processing."
+            log.info(msg)
```
gipert, line 332: remove

python/remage/cli.py
```diff
@@ -262,29 +314,84 @@ def remage_run_from_args(
...
+                [main_output_file] if py_args.merge_output_files else remage_files
+            )
+
+            # merge output files if requested
+            if py_args.reshape_output or py_args.merge_output_files:
+                original_files = utils.make_tmp(remage_files)
+
+            # if reshaping was requested call reboost
+            if py_args.reshape_output:
+                msg = "Begin reshaping output files."
```
gipert, line 346:
```suggestion
                msg = "Reshaping output files"
```

python/remage/cli.py
```diff
@@ -262,29 +314,84 @@ def remage_run_from_args(
...
+                    output=main_output_file,
+                    overwrite=overwrite_output,
+                )
+
+            if py_args.merge_output_files or py_args.reshape_output:
+                # set the merged output file for downstream consumers.
+                ipc_info.set("output", output_files)
+
+                msg = "Deleting original remage files."
+                log.info(msg)
```
gipert, line 385: remove

gipert:
please resolve all conversation, then good to merge from my side

pyproject.toml
tdixon97: I chose this since I think the latest `legend-pydataobj` has quite some breaking changes?

gipert: fixing versions is not that nice. we should support the latest version of lgdo and only set lower bounds

tdixon97: sure but then lets see how much the changes to LGDO break ... my guess is a lot

gipert: i don't think so

gipert: anyways, you will still not get pre-releases with the lower bound. just the latest stable

ManuelHu: there is no sane way to support iterating lh5 files (tables etc.) compatible with the old and new version... So yes, it will break a lot - not in here in remage, but in reboost.

mainly this API change:
https://github.com/legend-exp/legend-pydataobj/compare/v1.11.7...v1.12.0a3#diff-faf5b5aff0920778b3bced345801fb7b75cf06bd2866c36f792ea39c7ef17318L498-R519

tdixon97:
This should now be working, made some tests which work fine. More serious testing on the post-processing methods themself should be done in reboost.

tdixon97:
@gipert I don't get the read the docs failure

pyproject.toml
gipert: we should then start working on updating reboost soon... @tdixon97  in the meanwhile you can set >??? and <1.12 or so here

gipert:
not our fault, seems some random failure

pyproject.toml
gipert: and open an issue about removing <1.12 or we will forget

tdixon97:
I would merge this then

ManuelHu:
there is still the open discussion to the version bound. please set to < 0.12; to get fixes if somebody still releases new 0.11.x versions after 0.11.7....

PR #344, opened 2025-04-29
standardize macro commands

gipert:
- allow for optional arguments with boolean commands
- remove "Set" prefixes


src/RMGVertexConfinement.cc
```diff
@@ -957,12 +957,14 @@ void RMGVertexConfinement::DefineCommands() {
       ->DeclareMethod("Reset", &RMGVertexConfinement::Reset)
       .SetGuidance("Reset all parameters of vertex confinement, so that it can be reconfigured.")
       .SetStates(G4State_PreInit, G4State_Idle)
+      .SetDefaultValue("true")
```
ManuelHu, line 960: also no parameter?

gipert:
@legend-exp/remage-software please have a look at what I did here. This is the right way to define boolean commands such that one can omit the true/false argument, in case the command name is already self-explanatory.

gipert:
@ManuelHu @MoritzNeuberger is the Ci failure due to that warning about the very long time?

ManuelHu:
yes

PR #346, opened 2025-04-30
remove realms functionality

gipert:
Leave in the default cuts, for now.

### DIFF
--- a/src/RMGPhysics.cc
+++ b/src/RMGPhysics.cc
@@ -78,24 +78,37 @@ namespace u = CLHEP;
 
 RMGPhysics::RMGPhysics() {
 
-  G4VUserPhysicsList::defaultCutValue = 0.1 * u::mm;
-  this->SetPhysicsRealm(PhysicsRealm::kDoubleBetaDecay);
-
+  // verbosity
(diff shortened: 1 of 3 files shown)


src/RMGPhysics.cc
```diff
@@ -79,7 +79,6 @@ namespace u = CLHEP;
 RMGPhysics::RMGPhysics() {
 
   G4VUserPhysicsList::defaultCutValue = 0.1 * u::mm;
```
ManuelHu, line 81:
```suggestion
  G4VUserPhysicsList::defaultCutValue = 0.1 * u::mm;
  this->SetCuts();
```

otherwise SetCuts is not called by default...

tdixon97:
Overall looks good, I think we should consider changing some defaults.
- turn on angular correlations by default (people don't check these things),
- turn on the atomic physics PIXE etc,

tdixon97 (approved):

gipert:
@tdixon97 please add this to #43

gipert:
there's still something wrong, the ci is taking ages. i will have a closer look tomorrow

ManuelHu (approved):

ManuelHu:
the CI failures are not related to these changes, and are already fixed on main

PR #348, opened 2025-04-30
use DYLD_FALLBACK_LIBRARY_PATH for macOS

ManuelHu:


ManuelHu:
this cannot be easily fixed. The problem is - hopefully as simple as it can get:

* geant4 links against some native libraries (including libicu) that are all managed with conda
* uv in the remage build process does not communicate/use conda (of course; as expected, why would it?)
* the typical python package wheels are built against system libraries shipped with macOS, or ship their own dependencies
* the macOS dependencies are not ABI-compatible with the conda-installed ones
* python packages try to use their own libraries, but find conda ones...

example of the numpy failure:
* numpy links against accelerate, a BLAS implementation from the system
* conda constains openBLAS, another BLAS implementation
* when openBLAS is in DYLD_LIBRARY_PATH, then the dynamic linker gets confused

this could be technically cirvumvented. More concerning is the chain
* geant4 -> xerces-c -> icu which is definitely an important dependency (read: for GDML reading)
* but python packages expect the system libicu and bail if the conda-installed one is present

## all in all, it is a massive mess

PR #359, opened 2025-05-02
output: do not create root file by default

### LINKED ISSUE #326: default output file
currently, if no output file is specified, remage creates a file named `detector-hits.root` (this has been the case since at least 2022...)

I do not like this, for multiple reasons:
* we don't want to use root
* there is no way to disable output when detectors are registered
* it is unexpected for users, the default file name would nearly always be wrong. It is also not documented?

There are multiple ways to fix this:
1. do not write an output file by default
    a) error when trying to register an output scheme (i.e. to prevent data loss)
    b) silently discard output, maybe emitting a warning?
2. switch to default lh5 output (I still don't like that), and documednt it

ManuelHu:
fixes #326

### DIFF
--- a/src/RMGRunAction.cc
+++ b/src/RMGRunAction.cc
@@ -93,6 +93,11 @@ void RMGRunAction::BeginOfRunAction(const G4Run*) {
 
   if (fIsPersistencyEnabled) this->SetupAnalysisManager();
 
+  if (!manager->HasOutputFileName()) {
+    manager->EnablePersistency(false);
+    fIsPersistencyEnabled = false;
+  }
+
   // Check again, SetupAnalysisManager might have modified fIsPersistencyEnabled.
   if (fIsPersistencyEnabled) {
     auto ana_man = G4AnalysisManager::Instance();
@@ -138,9 +143,11 @@ void RMGRunAction::BeginOfRunAction(const G4Run*) {
 
   if (!fIsPersistencyEnabled && this->IsMaster()) {
     // Warn user if persistency is disabled if there are detectors defined.
-    auto level = manager->GetDetectorConstruction()->GetAllActiveOutputSchemes().empty()
-                     ? RMGLog::summary
-                     : RMGLog::warning;
+    auto level = RMGLog::summary;
+    if (!manager->GetDetectorConstruction()->GetAllActiveOutputSchemes().empty() &&
+        !manager->HasOutputFileNameNone()) {
+      level = RMGLog::warning;
+    }
     RMGLog::Out(level, "Object persistency disabled");
   }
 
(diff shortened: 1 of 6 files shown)


tdixon97:
Stop working
From: Manuel Huber ***@***.***>
Sent: Friday, May 2, 2025 8:48 PM
To: legend-exp/remage ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [legend-exp/remage] output: do not create root file by default (PR #359)


⚠ Caution: External sender


fixes #326<https://github.com/legend-exp/remage/issues/326>

You can view, comment on, or merge this pull request online at:

  https://github.com/legend-exp/remage/pull/359

Commit Summary

  *   03c2541<https://github.com/legend-exp/remage/pull/359/commits/03c25414fbe799437443e007c29d9a554cc19f22> output: do not create root file by default

File Changes

(2 files<https://github.com/legend-exp/remage/pull/359/files>)

  *   M include/RMGManager.hh<https://github.com/legend-exp/remage/pull/359/files#diff-3b9259d28c3e5b88e32a53169fa78100c6de7cf7fbd81f5792f79c7f98516510> (2)
  *   M src/RMGRunAction.cc<https://github.com/legend-exp/remage/pull/359/files#diff-ae774f976d003a4299d3e5f8260183e03ac90da5176e8f404a649102175fade0> (7)

Patch Links:

  *   https://github.com/legend-exp/remage/pull/359.patch
  *   https://github.com/legend-exp/remage/pull/359.diff

src/RMGRunAction.cc
```diff
@@ -140,7 +145,7 @@ void RMGRunAction::BeginOfRunAction(const G4Run*) {
     // Warn user if persistency is disabled if there are detectors defined.
     auto level = manager->GetDetectorConstruction()->GetAllActiveOutputSchemes().empty()
                      ? RMGLog::summary
-                     : RMGLog::warning;
+                     : RMGLog::error;
```
gipert, line 148: uhm are you sure? i don't have a strong option but i maybe feel this should be a warning

ManuelHu: you mean leaving it as-is? I am fine with that, that would also be my personal choice. In #326 it was just requested this will get an error....

tdixon97:
I wouldnt give an error

tdixon97:
It could be important for testing how long operations within the output scheme take

ManuelHu:
Okay, so I will change this at some point.

For the failing tests, shoulw I just add a dummy output file for these? Or should we design a way to instruct remage to _knowingly_ produce no output file?

gipert:
a CLI option like `--no-output-file`?

ManuelHu:
I would have used `-o none` - this is not a valid file name anyway, as Geant4 always needs to have a file extension

gipert:
sounds good!

tdixon97:
I think it shouldn't be needed to give a flag

ManuelHu:
No, the flag essentially only disables the warning

tdixon97:
Ah just for testing?

tdixon97:
I think the warning can be info instead

gipert:
no i like the warning... `-o none` would be interpreted as "i know what i'm doing and i want to suppress the warning"

tdixon97:
Why do you need a warning to say you dont get an output file if yoi never asked one, its the logical behaviour

ManuelHu:
But you also have specifically asked for output (i.e. by registering output schemes for detectors) in the first place, so that the warning appears at all. So what is the logical behaviour when "asking for output (tables)" and "not asking for output (a file)" at the same time?

tdixon97:
We say output schemes can also be filters? I guess the application is just testing performance

ManuelHu:
I now also added some docs

PR #362, opened 2025-05-06
Fixes to LH5 post-processing code

gipert:
- implement context manager to deal with temporary files
- code cosmetics

### DIFF
--- a/python/remage/cli.py
+++ b/python/remage/cli.py
@@ -21,8 +21,6 @@ from . import utils
 from .find_remage import find_remage_cpp
 from .ipc import IpcResult, ipc_thread_fn
 
-log = logging.getLogger(__name__)
-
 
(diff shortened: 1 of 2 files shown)


python/remage/cli.py
```diff
@@ -269,15 +270,18 @@ def remage_run_from_args(
         type=float,
         metavar="",
         default=10,
-        help="Time window in microseconds to group steps for reshaping.",
+        help=(
+            "Time window (in microseconds) to use to group steps into events "
```
tdixon97, line 274: I am not sure id call this events

gipert: needs to be something intelligible for non-expert users. suggest an alternative

tdixon97: hit

PR #365, opened 2025-05-06
Add more generator docs

EricMEsch:
I just realized that @gipert also heavily changed the `generators.md` file in his recent PR. We have to figure out the conflict, but should be fairly simple as it is just docs.

I just went over the basics of each of the generators and linked to where to find more information.

Especially the MUSUN one is pretty annoying, as in the paper they state that if you want to get access to MUSUN you should contact him per mail... 

But i think anyways at some point we can just use `revertex` or something to convert the musun text files to `lh5` and get rid of this generator and use the `FromFile` instead?

### DIFF
--- a/docs/manual/generators.md
+++ b/docs/manual/generators.md
@@ -4,7 +4,6 @@
 
 :::{todo}
 
-- list of built-in generators and documentation
 - muon generators
   :::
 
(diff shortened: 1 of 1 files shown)


gipert:
thanks eric, i can include it in my pr

EricMEsch:
I can close this then i guess?^^

PR #367, opened 2025-05-07
confinement: fix surface sampling on native volumes

ManuelHu:
this should fix the problem of surface sampling not working with natively sampleable volumes. It has been forcing the use of the non-native sampling scheme, but was not properly checking that all parameters for this were configured.

other changes:
* partially revert #340 (a default value changed that was always overriden)
* move the warning for missing specified MaxNumberOfIntersections - this might be helpful if we can reach this somehow differently in the future.

please review @tdixon97

### DIFF
--- a/src/RMGVertexConfinement.cc
+++ b/src/RMGVertexConfinement.cc
@@ -229,7 +229,7 @@ void RMGVertexConfinement::SampleableObject::GetDirection(G4ThreeVector& dir, G4
 bool RMGVertexConfinement::SampleableObject::GenerateSurfacePoint(
     G4ThreeVector& vertex,
     size_t max_attempts,
-    size_t n_max
+    size_t max_intersections
 ) const {
 
   size_t calls = 0;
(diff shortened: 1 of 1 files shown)


tdixon97 (approved):
It looks ok, but I think setting a default value for `fSurfaceSampleMaxIntersections` is not a bad idea, since in most cases it cannot be say > 10. And this avoids people putting the macro command in 99% of cases

ManuelHu:
yeah, but that would be the wrong place for this

tdixon97:
yeah lets add a default for `fSurfaceSampleMaxIntersections` ?

ManuelHu:
Is there a warning when a user has a _too small_ setting anywhere? Or would this be biasing the result?

tdixon97:
Hmm yeah there should be, if you ever get N Intersection > Max Intersection we should give an error (I hope its implemented)

tdixon97 (approved):

PR #368, opened 2025-05-07
update the default pars

tdixon97:
Justifications [[here]](https://docs.google.com/presentation/d/1FgzLuQQN1ej4WSvdsH1eEcNkhRBq3BEnGdfQCcV2w4s/edit?slide=id.g355a9320ae0_0_176#slide=id.g355a9320ae0_0_176)

* Turn on pre-cluster by default
* Use 50 um for Germanium and 500 um for Scintillator
* Combined low energy e- tracks and redistribute gamma energy
* Keep production cuts at their current default
* Update the docs
* Change the lh5 table to be `stp` also in reshaped files

### DIFF
--- a/docs/manual/output.md
+++ b/docs/manual/output.md
@@ -243,13 +243,15 @@ pointer to a new collection of clustered hits.
   similar function just needs to be written.
   :::
 
-Pre-clustering can be enabled with
+Pre-clustering is enabled by default for the Scintillator and Germanium
+output schemes, it can be disabled with the command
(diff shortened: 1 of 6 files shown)


gipert:
need to change the cli flags too

tdixon97:
That will require changing most tests. I would like to work a bit on reboost to handle reshaped outputs before doing that by default

gipert:
I insist that we should have also reshaping on by default in the next release. the test updates are minimal, I can work on it now

PR #371, opened 2025-05-08
docs: let prettier enforce line lengths in markdown

ManuelHu:
and some other fixes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -78,6 +78,11 @@ repos:
         language: pygrep
         entry: PyBind|Numpy|Cmake|CCache|Github|PyTest
         exclude: .pre-commit-config.yaml
+      - id: disallow-end-fence
+        name: "disallow ::: fences not on their own line"
+        language: pygrep
+        entry: "[^\\n]:*:::\\n|[^\\n]:*:::\\{"
+        types_or: [markdown]
 
   - repo: https://github.com/abravalheri/validate-pyproject
     rev: "v0.24.1"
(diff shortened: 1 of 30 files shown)


.pre-commit-config.yaml
```diff
@@ -39,7 +39,10 @@ repos:
     rev: "v3.5.3"
     hooks:
       - id: prettier
-        types_or: [yaml, markdown, html, css, scss, javascript, json]
+        types_or: [yaml, html, css, scss, javascript, json]
+      - id: prettier
+        types_or: [markdown]
+        args: ["--prose-wrap=always"]
```
gipert, line 45: the problem is that we really need to be careful with the spacing in `:::` blocks... this is why i removed it at some point. we can try to be careful

ManuelHu: we could also add a pygrep hook to catch `:::` at the end of lines, but not not on a line itself?

the start tags get ugly when not on their own line, but should typically still work. Only the end tags work only on their own line.

gipert: yeah maybe, but i'm fine with merging this for now

ManuelHu:
I will still add some more docs fixes that I found.

Notably, the build failure is not because of these changes; it also happens on main

ManuelHu:
should we start pinning older versions of snowballstemmer (they released v3 some hours ago, which broke everything); or should we wait to let it settle?

Problem: in the meantime we cannot reaonasbly publish a new release, as the rtd build will also be broken.

ManuelHu:
okay, I will add it to this PR

ManuelHu:
> maybe let's pin it and open an issue about it, we want to release soon...

@gipert the version pinning is not necessary any more, the broken version has been yanked on pypi for now.

PR #379, opened 2025-05-10
use a FNV-1a hash to compute process indices

ManuelHu:


tdixon97:
Looks great, I can work on the updates to reboost to also merge these tables. I guess we also need an update to `lh5concat` to not end up with the processes table twice

ManuelHu:
updated to not get negative process ids

Issue #385, opened 2025-05-20
Seeding with bxdecay0


tdixon97:
Currently seeds are hardcoded in the macro command, we should find a way to use the internal seeds of remage

ManuelHu:
probably not without patching bxdecay0, as we would need to change the definition of the macro command?

ManuelHu:
probably we could set our own seed before the run in C++ code, but the then-useless parameter would remain in the macro commands?

gipert:
Why can't we pass a seed to the bxdecay0 generator action? We had a quick look at the bxdecay0 source code with tobi and it might be possible, through this "configuration" object.

ManuelHu:
that is what I meant with
> probably we could set our own seed before the run in C++ code

but the (then) useless parameter would still remain in the messenger, no?

gipert:
I don't understand what you mean... We should grab the seed (or whatever manipulation of numbers coming out of the G4 random engine) and pass it directly to the bxdecay0 generator action, and this would not involve using any macro command or messenger class

ManuelHu:
but the user still has to set up the generator with `/bxdecay0/generator/dbd ISOTOPE SEED MODE LEVEL`, which still reuqires a seed that would then be ignored?

EricMEsch:
So wrapping bxdecay in remage works perfectly fine. 
Right now you do have to specify:
```console
/RMG/Generator/Select BxDecay0
```
I would then go on and add macro commands to RMGGeneratorDecay0 where you can specify stuff.

Question: Do you just want to wrap the cases we are interested in? So just have one macro like
```console
/RMG/Generator/Decay0 2vbb
```
And the rest is handled by remage?
Because the 4 variables i set are
```c++
G4String nuclide;
G4int seed;
G4int dbd_mode;
G4int dbd_level;
```
But if I just forward them what is the gain?

Another question: Do we entirely want to disable the `/bxdecay0/generator/dbd ...` command? I think not? As right now if you do not specify the command remage will set its own defaults. And if you specify it, it will overwrite anything set by remage.
@gipert @tdixon97

PR #390, opened 2025-05-22
Wrap 2vbb and 0vbb in remage

EricMEsch:
So this adds the simple macro command `/RMG/Generator/Decay0/Mode` to remage with the two options `2vbb` and `0vbb`.

These options can be expanded easily. (Or change the name to `/RMG/Generator/Decay/Mode` if thats better)

The simulation will always use the last specified command. So in case of:
```console
/RMG/Generator/Decay0/Mode 2vbb
/bxdecay0/generator/dbd Xe136 2345 4 0
```
Xe-136 will be simulated and vice versa.
If you want some tests on this for the CI i am not sure, anyone has ideas what tests we would like?

### DIFF
--- a/include/RMGGeneratorDecay0.hh
+++ b/include/RMGGeneratorDecay0.hh
@@ -42,28 +42,33 @@ class RMGGeneratorDecay0 : public RMGVGenerator {
 
   public:
 
-    // This matches the BxDecay0 numbering, shifted by 1 as BxDecay0 starts counting at 1.
+    /**
+     * @enum Process
+     * @brief Enumeration of nuclear decay processes supported by the generator.
+     *
+     * This matches the BxDecay0 numbering, shifted by 1 as BxDecay0 starts counting at 1.
+     */
     enum class Process {
-      k0vbb,           // neutrinoless double beta decay
-      k0vbb_lambda_0,  // 0+ -> 0+ {2n} with RHC lambda
-      k0vbb_lambda_02, // 0+ -> 0+, 2+ {N*} with RHC lambda
-      k2vbb,           // 2 neutrino double beta decay
-      k0vbb_M1,        // 0+ -> 0+ {2n} (Majoron, SI=1)
-      k0vbb_M2,        // 0+ -> 0+ {2n} (Majoron, SI=2)
-      k0vbb_M3,        // 0+ -> 0+ {2n} (Majoron, SI=3)
-      k0vbb_M7,        // 0+ -> 0+ {2n} (Majoron, SI=7)
-      k0vbb_lambda_2,  // 0+ -> 2+ {2n} with RHC lambda
-      k2vbb_2,         // 0+ -> 2+ {2n}, {N*}
-      k0vkb,           // EC + beta+  0+ -> 0+, 2+
-      k2vkb,           // EC + beta+  0+ -> 0+, 2+
-      k0v2k,           // double EC  0+ -> 0+, 2+
-      k2v2k,           // double EC  0+ -> 0+, 2+
-      k2vbb_bos0,      //  0+ -> 0+ with bosonic neutrinos
-      k2vbb_bos2,      // 0+ -> 2+ with bosonic neutrinos
-      k0vbb_eta_s,     // 0+ -> 0+ with RHC eta simplified expression
-      k0vbb_eta_nmes,  //  0+ -> 0+ with RHC eta and specific NMEs
-      k2vbb_lv,        // 0+ -> 0+ with Lorentz violation
-      k0v4b            // 0+ -> 0+ Quadruple beta decay
+      /** @brief neutrinoless double beta decay */ k0vbb = 1,
+      /** @brief 0+ -> 0+ {2n} with RHC lambda */ k0vbb_lambda_0 = 2,
+      /** @brief 0+ -> 0+, 2+ {N*} with RHC lambda */ k0vbb_lambda_02 = 3,
+      /** @brief 2 neutrino double beta decay */ k2vbb = 4,
+      /** @brief 0+ -> 0+ {2n} (Majoron, SI=1) */ k0vbb_M1 = 5,
+      /** @brief 0+ -> 0+ {2n} (Majoron, SI=2) */ k0vbb_M2 = 6,
+      /** @brief 0+ -> 0+ {2n} (Majoron, SI=3) */ k0vbb_M3 = 7,
+      /** @brief 0+ -> 0+ {2n} (Majoron, SI=7) */ k0vbb_M7 = 8,
+      /** @brief 0+ -> 2+ {2n} with RHC lambda */ k0vbb_lambda_2 = 9,
+      /** @brief 0+ -> 2+ {2n}, {N*} */ k2vbb_2 = 10,
+      /** @brief EC + beta+  0+ -> 0+, 2+ */ k0vkb = 11,
+      /** @brief EC + beta+  0+ -> 0+, 2+ */ k2vkb = 12,
+      /** @brief double EC  0+ -> 0+, 2+ */ k0v2k = 13,
+      /** @brief double EC  0+ -> 0+, 2+ */ k2v2k = 14,
+      /** @brief 0+ -> 0+ with bosonic neutrinos */ k2vbb_bos0 = 15,
+      /** @brief 0+ -> 2+ with bosonic neutrinos */ k2vbb_bos2 = 16,
+      /** @brief 0+ -> 0+ with RHC eta simplified expression */ k0vbb_eta_s = 17,
+      /** @brief 0+ -> 0+ with RHC eta and specific NMEs */ k0vbb_eta_nmes = 18,
+      /** @brief 0+ -> 0+ with Lorentz violation */ k2vbb_lv = 19,
+      /** @brief 0+ -> 0+ Quadruple beta decay */ k0v4b = 20
     };
 
     /** @brief Constructor that links the BxDecay0 generator action to remage.
@@ -88,14 +93,12 @@ class RMGGeneratorDecay0 : public RMGVGenerator {
      *  @details Only does something if any remage set up command was used ( @c fUpdateSeeds is true).
      */
     void BeginOfRunAction(const G4Run*) override;
-    inline void EndOfRunAction(const G4Run*) override {}
+    void EndOfRunAction(const G4Run*) override {}
 
     /** @brief Sets BxDecay0 to run in background mode and sets the specific isotope.
      *  @param isotope The isotope to set (e.g. "Co60").
      */
     void SetBackground(std::string);
-
-    void SetUpdateSeeds(bool value) { fUpdateSeeds = value; }
 
 
   private:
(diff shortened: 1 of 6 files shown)


EricMEsch:
Fixes #385

src/RMGGeneratorDecay0.cc
```diff
@@ -43,4 +50,53 @@ void RMGGeneratorDecay0::GeneratePrimaries(G4Event* event) {
...
+  fDecay0G4Generator->SetConfigHasChanged(true);
+}
+
+void RMGGeneratorDecay0::DefineCommands() {
+
+  // NOTE: SetUnit(Category) is not thread-safe
+
+  fMessenger = std::make_unique<G4GenericMessenger>(
+      this,
+      "/RMG/Generator/Decay0/",
```
gipert, line 90:
```suggestion
      "/RMG/Generator/BxDecay0/",
```

src/RMGGeneratorDecay0.cc
```diff
@@ -43,4 +50,53 @@ void RMGGeneratorDecay0::GeneratePrimaries(G4Event* event) {
...
+      "Commands for controlling the BxDecay0 generator"
+  );
+
+  fMessenger->DeclareMethod("Mode", &RMGGeneratorDecay0::SetMode)
```
gipert, line 94: We could maybe call this 
```suggestion
  fMessenger->DeclareMethod("Process", &RMGGeneratorDecay0::SetMode)
```

ManuelHu:
Also, please rename the SetMode command and method to make clear that we are solely talking about 76-Ge here (maybe Ge76DecayMode or similar?)

Am 22. Mai 2025 15:32:10 MESZ schrieb Luigi Pertoldi ***@***.***>:
>@gipert commented on this pull request.
>
>
>
>> +void RMGGeneratorDecay0::SetMode(std::string mode) {
>+  try {
>+    fDecayMode = RMGTools::ToEnum<RMGGeneratorDecay0::DecayMode>(mode, "decay mode");
>+  } catch (const std::bad_cast&) { return; }
>+  if (fDecayMode == DecayMode::k2vbb) {
>+    nuclide = "Ge76";
>+    dbd_mode = 4;
>+    dbd_level = 0;
>+  } else if (fDecayMode == DecayMode::k0vbb) {
>+    nuclide = "Ge76";
>+    dbd_mode = 1;
>+    dbd_level = 0;
>+  } else {
>+    RMGLog::Out(RMGLog::error, "Unknown decay mode");
>+  }
>+  seed = static_cast<G4int>(G4UniformRand() * std::numeric_limits<G4int>::max());
>
>It's impossible to achieve reproducibility this way, what we want to do instead is to use the Geant4 seed (that might have been set by the user through the remage macro command). I don't know how to do this so you'll have to do some research
>

include/RMGGeneratorDecay0.hh
```diff
@@ -32,6 +33,11 @@ class RMGGeneratorDecay0 : public RMGVGenerator {
 
   public:
 
+    enum class DecayMode {
```
gipert, line 36:
```suggestion
    enum class Process {
```
?

ManuelHu:
but when you support all isotopes and all 19 processes, we also have to support the different energy levels? The case is just simple in our case as we know what we want with Ge76, but that is not very general?

gipert:
For sure we want to support everything about Ge76, processes and energy levels. And we might also want to simulate things like Ar36 2νECEC in the future...

ManuelHu:
I understand that @gipert also wants to replace the numeric mode selection so that it is not open to errors any more

include/RMGGeneratorDecay0.hh
EricMEsch: Well it is literaly called "DecayMode" in BxDecay

EricMEsch:
I have also just found out Geant4 does not like longs in a macro command it seems, because it expects an int:
image
This is the source code of G4GenericMessenger:
```c++
if (tInfo == typeid(int) || tInfo == typeid(long) || tInfo == typeid(unsigned int)
        || tInfo == typeid(unsigned long))
    {
      ptype = 'i';
    }
```

There is the class `G4UIcmdWithALongInt` but it can not even find any place it is actually used...
I would change the argument of `SetRandEngineSeed` to `int` if there are no objections

gipert:
Ok!

src/RMGGeneratorDecay0.cc
```diff
@@ -43,4 +50,75 @@ void RMGGeneratorDecay0::GeneratePrimaries(G4Event* event) {
...
+    dbd_level = 0;
+  } else {
+    RMGLog::Out(RMGLog::error, "Unknown decay mode");
+  }
+
+  seed = CLHEP::HepRandom::getTheSeed();
+
+  // Sanity check
+  if (seed <= 0) {
+    RMGLog::Out(RMGLog::error, "Seed seems invalid or uninitialized. Setting seed to 0.");
```
ManuelHu, line 73:
```suggestion
    RMGLog::Out(RMGLog::error, "CLHEP seed seems invalid or uninitialized. Setting BxDecay0 seed to 0.");
```

this should be more clear, you do not set the global seed here.

EricMEsch: You are right, i did actually already change it locally for the warning message :-)

src/RMGGeneratorDecay0.cc
```diff
@@ -43,4 +50,129 @@ void RMGGeneratorDecay0::GeneratePrimaries(G4Event* event) {
...
+
+
+  fGeneratorCmd = new G4UIcommand("/RMG/Generator/BxDecay0/dbd", this);
+  fGeneratorCmd
+      ->SetGuidance("Set the isotope, process and energy level for the double beta decay mode of the BxDecay0 generator");
+
+  auto isotope = new G4UIparameter("isotope", 's', false);
+  isotope->SetGuidance(
+      "Set the isotope for the double beta decay"
+  ); // Do we want to add an enum for every supported isotope?
```
ManuelHu, line 131:
```suggestion
  isotope->SetGuidance("Set the isotope for the double beta decay");
  std::string isot_cand;
  for (const auto& isot : bxdecay0::dbd_isotopes()) { isot_cand += std::string(isot) + ' '; }
  isotope->SetParameterCandidates(isot_cand.c_str());
```

bxdecay0 will complain (i.e., throw an exception) when it encounters an undefined isotope. However, that leads to a hang in MT mode as Geant4 doesn't like exceptions to be raised in MT mode before the run (yeah, I know...)

the candidates list should guard against that...

src/RMGGeneratorDecay0.cc
```diff
@@ -15,9 +15,15 @@
...
+
+#include "Randomize.hh"
+
 #include "RMGConfig.hh"
 #include "RMGLog.hh"
 #include "RMGManager.hh"
+#include "RMGTools.hh"
+
 #if RMG_HAS_BXDECAY0
 #include "bxdecay0_g4/primary_generator_action.hh"
```
ManuelHu, line 28:
```suggestion
#include "bxdecay0/bb_utils.h"
#include "bxdecay0_g4/primary_generator_action.hh"
```

(needed for a comment below)

src/RMGGeneratorDecay0.cc
```diff
@@ -43,4 +50,129 @@ void RMGGeneratorDecay0::GeneratePrimaries(G4Event* event) {
...
+}
+
+void RMGGeneratorDecay0::DefineCommands() {
+
+  fMessenger = std::make_unique<G4GenericMessenger>(
+      this,
+      "/RMG/Generator/BxDecay0/",
+      "Commands for controlling the BxDecay0 generator"
+  );
+
```
ManuelHu, line 109:
```suggestion
  std::string isot_cand;
  for (const auto& isot : bxdecay0::background_isotopes()) { isot_cand += std::string(isot) + ' '; }
```

src/RMGGeneratorDecay0.cc
```diff
@@ -43,4 +50,129 @@ void RMGGeneratorDecay0::GeneratePrimaries(G4Event* event) {
...
+  // Need an additional command because both modes expect different parameters.
+  fMessenger->DeclareMethod("background", &RMGGeneratorDecay0::SetBackground)
+      .SetGuidance("Set the isotope for the background mode of the BxDecay0 generator. E.g. 'Co60'")
+      .SetParameterName("isotope", false) // Do we want to add an enum for every supported isotope?
```
ManuelHu, line 113:
```suggestion
      .SetParameterName("isotope", false)
      .SetCandidates(isot_cand)
```

src/RMGGeneratorDecay0.cc
```diff
@@ -43,4 +50,129 @@ void RMGGeneratorDecay0::GeneratePrimaries(G4Event* event) {
...
+          RMGLog::error,
+          "CLHEP seed seems invalid or uninitialized. Setting BxDecay0 seed to 0."
+      );
+      seed = 0;
+    }
+
+    // This should never occur on most systems, as the only way is through the internal CLHEP table
+    // and the values there are all below the max int32.
+    if (seed > std::numeric_limits<int>::max()) {
+      int new_seed = seed - std::numeric_limits<int>::max();
```
ManuelHu, line 70:
```suggestion
      long new_seed = seed - std::numeric_limits<int>::max();
```
fixes a clang-tidy warning (we store it in a long variable anyway)

src/RMGGeneratorDecay0.cc
```diff
@@ -43,4 +50,129 @@ void RMGGeneratorDecay0::GeneratePrimaries(G4Event* event) {
...
+          std::numeric_limits<int>::max(),
+          ". Setting BxDecay0 seed to ",
+          new_seed
+      );
+      seed = new_seed;
+    }
+
+    bxdecay0_g4::PrimaryGeneratorAction::ConfigurationInterface&
+        configInt = fDecay0G4Generator->GrabConfiguration();
+    configInt.seed = seed;
```
ManuelHu, line 85:
```suggestion
    configInt.seed = static_cast<int>(seed);
```

include/RMGGeneratorDecay0.hh
```diff
@@ -41,12 +79,59 @@ class RMGGeneratorDecay0 : public RMGVGenerator {
...
+     */
+    void BeginOfRunAction(const G4Run*) override;
+    inline void EndOfRunAction(const G4Run*) override {}
+
+    /** @brief Sets BxDecay0 to run in background mode and sets the specific isotope.
+     *  @param isotope The isotope to set (e.g. "Co60").
+     */
+    void SetBackground(std::string);
+
+    void SetUpdateSeeds(bool value) { fUpdateSeeds = value; }
```
ManuelHu, line 98: do we need this function. This is only needed for resetting this flag from external code, but do we want this?

EricMEsch:
The only issue i see now if someone makes changes to macro commands without having the BxDecay0 module, then does `make remage-doc-dump` and pushes, the doc-dump will not include the `BxDecay0` commands anymore.

Which means the fact if they are included in the macro command list depends how who last pushed to remage...

tdixon97:
I just realised the user manual was not updated, can you please do this @EricMEsch ?

PR #396, opened 2025-06-05
Add BxDecay0 docs to the user manual

EricMEsch:
-rework bxdecay docs 
-add note to extend for updating from old versions where remants of old dependencies might cause issues
-add hadronic decay time threshold info to physicslist. If i have time i can also do some more work on the physicslist part of the usermanual, although i am honestly no expert in what the difference between them is. I can atleast explain what options there are and how to choose

### DIFF
--- a/docs/manual/extend.md
+++ b/docs/manual/extend.md
@@ -11,15 +11,6 @@ find_package(remage REQUIRED)
 target_link_libraries(myapp PRIVATE RMG::remage)
 ```
 
-:::{note}
-
-When updating from older _remage_ versions you might encounter some issues with
-remnants of old dependencies. In this case fully delete your current _remage_
-installation and re-install into a clean folder, as the install process itself
-will not remove any files that are not on the list of to-be-installed files.
-
-:::
-
 ## Forwarding arguments
 
 The {cpp:class}`RMGManager` is crucial for any of your code to extend _remage_.
--- a/docs/manual/extend.md
+++ b/docs/manual/extend.md
@@ -11,6 +11,15 @@ find_package(remage REQUIRED)
 target_link_libraries(myapp PRIVATE RMG::remage)
 ```
 
+:::{note}
+
+When updating from older _remage_ versions you might encounter some issues with
+remnants of old dependencies. In this case fully delete your current _remage_
+installation and re-install into a clean folder, as the install process itself
+will not remove any files that are not on the list of to-be-installed files.
+
+:::
+
 ## Forwarding arguments
 
 The {cpp:class}`RMGManager` is crucial for any of your code to extend _remage_.
--- a/docs/manual/generators.md
+++ b/docs/manual/generators.md
@@ -206,10 +206,16 @@ The second macro command is used to generate background physics with _bxdecay0_.
 The command <project:../rmg-commands.md#rmggeneratorbxdecay0background> takes
 only a single string as parameter, which should be the name of the isotope.
 Again all isotope candidates are listed
-[here](project:../rmg-commands.md#rmggeneratorbxdecay0background).
+[here](project:../rmg-commands.md#rmggeneratorbxdecay0background). Using these
+we can for example simulate a $^{60}$Co background source:
 
-_remage_ will forward the current seed of Geant4 to _bxdecay0_. This means that
-if any seed is specified using the
+```
+/RMG/Generator/Select BxDecay0
+/RMG/Generator/BxDecay0/background Co60
+```
+
+The current seed of Geant4 will be forwarded to _bxdecay0_. This means that if
+any seed is specified using the
 <project:../rmg-commands.md#rmgmanagerrandomization> commands, that seed will be
 forwarded to _bxdecay0_.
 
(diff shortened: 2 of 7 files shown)


docs/manual/extend.md
```diff
@@ -11,6 +11,15 @@ find_package(remage REQUIRED)
...
 ```
 
+:::{note}
+
+When updating from older _remage_ versions you might encounter some issues with
+remnants of old dependencies. In this case fully delete your current _remage_
+installation and re-install into a clean folder, as the install process itself
+will not remove any files that are not on the list of to-be-installed files.
+
+:::
```
gipert, line 21: I find hard to understand what this means, I would simplify by just saying that if there's any unexpected error one should always try to compile from scratch (i.e. remove the build folder). also, I would move this in install.md

docs/manual/generators.md
```diff
@@ -173,48 +173,65 @@ This mechanism can be used to select arbitrary sections of decay chains!
...
-The only documentation for this extension is available
+The main documentation for this extension is available
 [here](https://github.com/BxCppDev/bxdecay0).
 
 The macro [command](project:../rmg-commands.md#rmggeneratorselect):
 
 `/RMG/Generator/Select BxDecay0`
 
-can be used to select the Decay0 generators.
+can be used to select the Decay0 generator.
```
gipert, line 183:
```suggestion
can be used to select the BxDecay0 generator.
```

docs/manual/generators.md
```diff
@@ -173,48 +173,65 @@ This mechanism can be used to select arbitrary sections of decay chains!
...
-- `SEED` (int): is a random seed,
-- `MODE` (int): is an index the double beta decay mode labelling the decay type,
-  a table of implemented decay modes can be found in
-  [link](https://github.com/BxCppDev/bxdecay0/tree/develop?tab=readme-ov-file#list-of-supported-double-beta-decay-modes).
-- `LEVEL` (int): is a index of the energy level of the daughter nuclei, the list
-  of available levels are
-  [documented here](https://github.com/BxCppDev/bxdecay0/tree/develop?tab=readme-ov-file#list-of-daughter-nucleus-excited-states-in-double-beta-decay).
+The macro command to generate double beta decay physics
+<project:../rmg-commands.md#rmggeneratorbxdecay0dbd> expects the isotope name,
+the name of the process you want to simulate (like 2vbb or 0vbb) and an optional
```
gipert, line 192:
```suggestion
the name of the process you want to simulate (like `2vbb` or `0vbb`) and an optional
```

docs/manual/generators.md
```diff
@@ -173,48 +173,65 @@ This mechanism can be used to select arbitrary sections of decay chains!
...
+energy level for the daughter nucleus. All available isotopes and processes are
```
gipert, line 193:
```suggestion
level for the daughter nucleus. All available isotopes and processes are
```

docs/manual/generators.md
```diff
@@ -173,48 +173,65 @@ This mechanism can be used to select arbitrary sections of decay chains!
...
+listed [here](project:../rmg-commands.md#rmggeneratorbxdecay0dbd). The energy
+level defaults to zero and can therefore be omitted in most cases.
```
gipert, line 195:
```suggestion
listed [here](project:../rmg-commands.md#rmggeneratorbxdecay0dbd). The nuclear
level defaults to ground state and can therefore be omitted in most cases.
```

docs/manual/generators.md
```diff
@@ -173,48 +173,65 @@ This mechanism can be used to select arbitrary sections of decay chains!
...
 For example we can generate two-neutrino double beta decay to the $0^+$ ground
 state of $^{76}$Ge with:
 
 ```
-/bxdecay0/generator/dbd Ge76 1234 4 0
+/RMG/Generator/BxDecay0/dbd Ge76 2vbb
 ```
 
+The second macro command is used to generate background physics with _bxdecay0_.
```
gipert, line 205: "background" is a nebulous name, why not explicitly say "radioactive decay" or so?

docs/manual/generators.md
```diff
@@ -173,48 +173,65 @@ This mechanism can be used to select arbitrary sections of decay chains!
...
+/RMG/Generator/Select BxDecay0
+/RMG/Generator/BxDecay0/background Co60
+```
+
+The current seed of Geant4 will be forwarded to _bxdecay0_. This means that if
+any seed is specified using the
+<project:../rmg-commands.md#rmgmanagerrandomization> commands, that seed will be
+forwarded to _bxdecay0_.
+
+:::{note}
```
gipert, line 222:
```suggestion
:::{warning}
```

tdixon97:
Do we really want to support wrapping those commands, imo they are not implemented better than Geant4. If someone wants it they still have the Bxdecay0 commands

gipert:
@EricMEsch could you resolve the conversations?

I would also tend to think that we should discourage users to use BxDecay0's background generator... So maybe good to remove.

Also, I just realized, `dbd` is quite a bad name and deviates from the remage style. Can we use `DoubleBetaDecay`?

PR #398, opened 2025-06-09
Rename vertex ntuple to vtx

gipert:
Rename the primary vertex output table from `vertices` to `vtx` and move it one group back (LH5 output only)


gipert:
Waiting for this: https://github.com/legend-exp/reboost/pull/83

PR #412, opened 2025-06-17
Add docstrings to IPC helper functions

gipert:
## Summary
- document the IPC parsing logic and thread handling
- document `IpcResult` helper methods

## Testing
- `ruff check python/remage/ipc.py`
- `pytest -q` *(fails: ModuleNotFoundError: No module named 'numpy')*

------
https://chatgpt.com/codex/tasks/task_e_6851744647c083308079d015d10b6c41

### DIFF
--- a/python/remage/ipc.py
+++ b/python/remage/ipc.py
@@ -17,11 +17,28 @@ Message parts are separated using ASCII control characters. Each message ends wi
 indicate that the C++ process expects an acknowledgement with a POSIX signal before
 continuing.
 
+Each message must contain at least two records. The first is treated as the
+message's *key*, whereas the second one is the associated value:
+
++-------+--------+---------+-----------+--------+
+| *key* | ``RS`` | *value* | [``ENQ``] | ``GS`` |
++-------+--------+---------+-----------+--------+
+
 Records within a message are delimited by ``RS`` (record separator) and each record
 may contain multiple units split by ``US`` (unit separator). Records with more then
 one unit are returned as tuples on the python side.
 
-Each message must contain at least one record, it is treated as the message's *key*.
+Example: A message
+
++-------+--------+--------+--------+--------+--------+--------+-----------+--------+
+| *key* | ``RS`` | value0 | ``RS`` | value1 | ``US`` | value2 | [``ENQ``] | ``GS`` |
++-------+--------+--------+--------+--------+--------+--------+-----------+--------+
+
+would be decoded to
+
+```python
+["value0", ("value1", "value2")]
+```
 
 Blocking messages
 -----------------
--- a/python/remage/ipc.py
+++ b/python/remage/ipc.py
@@ -84,7 +84,7 @@ def ipc_thread_fn(
     each complete IPC message to :func:`handle_ipc_message` for parsing and
     handling of the associated action. Any message parts will be decoded as
     UTF-8 before parsing.
-    
+
     Blocking messages are acknowledged by sending the POSIX signal ``SIGUSR2``
     to the child  process, while fatal messages trigger ``SIGTERM``. Any
     messages that are not handled by :func:`handle_ipc_message` are appended
--- a/python/remage/ipc.py
+++ b/python/remage/ipc.py
@@ -3,7 +3,7 @@ IPC message receiver implementation for ``remage-cpp``.
 
 .. note ::
 
-    The C++ IPC *sender* implementation can be found in {cpp:class}`RMGIpc`.
+    The C++ IPC *sender* implementation can be found in :cpp:class:`RMGIpc`.
 
 
 Binary message format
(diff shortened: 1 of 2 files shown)


python/remage/ipc.py
```diff
@@ -42,6 +63,24 @@ def handle_ipc_message(msg: bytes) -> tuple[bool, list, bool]:
...
+    later processing.
+
+    Parameters
+    ----------
+    pipe_r
+        File descriptor for the read end of the IPC pipe.
+    proc
+        The subprocess running ``remage-cpp``.
+    unhandled_ipc_messages
+        List that receives messages which were not interpreted here.
```
ManuelHu, line 82:
```suggestion
        List that will receive messages whichg were not directly handled (i.e. for further processing)
```

python/remage/ipc.py
```diff
@@ -92,16 +134,20 @@ def get(self, name: str, expected_len: int = 1) -> list[str]:
         return msgs
 
     def get_single(self, name: str, default: str) -> str:
+        """Return a single value for ``name`` or ``default`` if not present."""
```
ManuelHu, line 137:
```suggestion
        """Return the single single value for the key ``name`` or ``default`` if not present.

        .. note ::
            if more then one value for the key had been submitted, this function will throw.
        """
```

python/remage/ipc.py
```diff
@@ -92,16 +134,20 @@ def get(self, name: str, expected_len: int = 1) -> list[str]:
...
         gen = self.get(name)
         if len(gen) > 1:
             msg = f"ipc returned key {name} more than once"
             raise RuntimeError(msg)
         return gen[0] if len(gen) == 1 else default
 
     def set(self, name: str, values: list[str]) -> None:
+        """Replace existing entries for ``name`` with ``values``."""
```
ManuelHu, line 145:
```suggestion
        """Replace existing stored messages for they key ``name`` with ``values``.

        .. note ::
            This will create a single-record message for each value in `values`. Setting more
            complex messages is not implemented yet.
        """
```

python/remage/ipc.py
```diff
@@ -92,16 +134,20 @@ def get(self, name: str, expected_len: int = 1) -> list[str]:
...
         self.remove(name)
         for v in values:
             self.ipc_info.append([name, v])
 
     def remove(self, name: str) -> None:
+        """Remove all records with the given ``name`` from the IPC info."""
+
```
ManuelHu, line 152:
```suggestion
        """Remove all messages with the given key ``name`` from the IPC info."""
```

python/remage/ipc.py
```diff
@@ -79,9 +118,12 @@ def ipc_thread_fn(
 
 class IpcResult:
     def __init__(self, ipc_info):
+        """Wrapper around the raw IPC information returned by ``remage-cpp``."""
+
         self.ipc_info = ipc_info
 
     def get(self, name: str, expected_len: int = 1) -> list[str]:
+        """Return all values of a given key from the IPC message list."""
```
ManuelHu, line 126:
```suggestion
        """Return all messages of a given key ``name`` from the IPC message list.

        Params
        ------
        expected_len
            only return messages that have the expected number of records. Other messages are
            silently skipped.
        """
```

python/remage/ipc.py
```diff
@@ -79,9 +118,12 @@ def ipc_thread_fn(
 
 class IpcResult:
     def __init__(self, ipc_info):
+        """Wrapper around the raw IPC information returned by ``remage-cpp``."""
```
ManuelHu, line 121:
```suggestion
        """Storage structure for the IPC messages returned by ``remage-cpp``."""
```

python/remage/ipc.py
```diff
@@ -11,7 +11,28 @@
...
 def handle_ipc_message(msg: bytes) -> tuple[bool, list, bool]:
-    # parse the IPC message structure.
+    """Parse a raw IPC message from ``remage-cpp``.
+
+    Messages are separated using ASCII control characters.  Each message ends
+    with ``GS`` (group separator) and may additionally end in ``ENQ`` (enquiry)
+    to indicate that the C++ process expects a response before continuing.
+    Records within a message are delimited by ``RS`` (record separator) and each
+    record may contain multiple units split by ``US`` (unit separator).  Units
+    beyond the first are returned as tuples.
```
ManuelHu, line 21:
```suggestion
    Messages are decoded as UTF-8 strings; transmitting binary (non-string) data
    with this IPC mechanism is not possible.

    Message parts are separated using ASCII control characters. Each message ends
    with ``GS`` (group separator) which may be additionally followed by ``ENQ``
    (enquiry) to indicate that the C++ process expects an acknowledgement with a
    POSIX signal before continuing.

    Records within a message are delimited by ``RS`` (record separator) and each
    record may contain multiple units split by ``US`` (unit separator). Records
    with more then one unit are returned as tuples.

    This function should directly handle all known blocking IPC messages, which
    will not be returned for subsequent processing.
```

python/remage/ipc.py
```diff
@@ -42,6 +63,24 @@ def handle_ipc_message(msg: bytes) -> tuple[bool, list, bool]:
...
     pipe_r: int, proc: subprocess.Popen, unhandled_ipc_messages: list
 ) -> None:
+    """Read and handle IPC messages coming from ``remage-cpp``.
+
+    This function runs in a dedicated thread.  It reads from ``pipe_r`` and
+    dispatches each complete IPC message to :func:`handle_ipc_message`.  Blocking
+    messages are acknowledged by sending ``SIGUSR2`` to the child process, while
+    fatal messages trigger ``SIGTERM``.  Any messages that are not handled by
+    :func:`handle_ipc_message` are appended to ``unhandled_ipc_messages`` for
+    later processing.
```
ManuelHu, line 73:
```suggestion
    .. important ::
        This function runs in a dedicated thread in
        :func:`remage.cli._run_remage_cpp` and should not be called directly by
        the user. The parameter ``unhandled_ipc_messages`` acts as a return
        value, as thread functions cannot directly return.


    The function reads from the pipe file descriptor ``pipe_r`` and dispatches
    each complete IPC message to :func:`handle_ipc_message` for parsing and
    handling of the associated action. Any message parts will be decoded as
    UTF-8 before parsing.
    
    Blocking messages are acknowledged by sending the POSIX signal ``SIGUSR2``
    to the child  process, while fatal messages trigger ``SIGTERM``. Any
    messages that are not handled by :func:`handle_ipc_message` are appended
    to the list ``unhandled_ipc_messages`` for later processing.
```

python/remage/ipc.py
```diff
@@ -11,7 +11,28 @@
 
 
 def handle_ipc_message(msg: bytes) -> tuple[bool, list, bool]:
```
ManuelHu, line 13:
```suggestion
def handle_ipc_message(msg: str) -> tuple[bool, list, bool]:
```

yeah, actually it already was converted to str below...

python/remage/ipc.py
```diff
@@ -11,7 +11,28 @@
...
+    with ``GS`` (group separator) and may additionally end in ``ENQ`` (enquiry)
+    to indicate that the C++ process expects a response before continuing.
+    Records within a message are delimited by ``RS`` (record separator) and each
+    record may contain multiple units split by ``US`` (unit separator).  Units
+    beyond the first are returned as tuples.
+
+    Parameters
+    ----------
+    msg
+        The raw message bytes including the trailing separator.
```
ManuelHu, line 26:
```suggestion
        The raw message bytes decoded to UTF-8, still including the trailing separator.
```

PR #413, opened 2025-06-18
correct the overwrite arg to reboost

tdixon97:


gipert:
added in https://github.com/legend-exp/remage/pull/408/commits/46c8c7ff220d3630f81bbf8f38b6967cbb70ea65

PR #418, opened 2025-06-23
More robust output format communication between remage-cpp RMGConvertLH5 and the wrapper

gipert:
- [x] handle non-stp auxiliary ntuples automatically when converting to LH5
- [x] better IPC messages for output table names
- [ ] include aux/non-aux flag info in IPC messages to wrapper
- [ ] ConvertFromLH5?

@ManuelHu this still needs work but maybe you can already have a look.


src/RMGRunAction.cc
```diff
@@ -82,9 +82,8 @@ void RMGRunAction::SetupAnalysisManager() {
   if (RMGLog::GetLogLevel() <= RMGLog::debug) ana_man->SetVerboseLevel(10);
   else ana_man->SetVerboseLevel(0);
 
-  // do it only for activated detectors (have to ask to the manager)
+  // do it only for activated detectors (karen wants to speak to the manager)
```
ManuelHu, line 85: ?

src/remage-from-lh5.cc
```diff
@@ -28,19 +28,31 @@ int main(int argc, char** argv) {
   bool verbosity = false;
   bool dry_run = false;
   std::vector<std::string> file_names;
+  std::set<std::string> aux_ntuples = {}; // not implemented
```
ManuelHu, line 31: probably not needed for now; remage-from-lh5 is only used for input files for the vertex reader. I do not ethink that we need this for now there.

src/RMGConvertLH5.cc
```diff
@@ -299,20 +299,28 @@ bool RMGConvertLH5::ConvertToLH5Internal() {
   auto ntuples_group = hfile.openGroup(ntuple_group_name);
   auto ntuples = GetChildren(ntuples_group);
   bool ntuple_success = true;
+  std::cout << std::endl;
```
ManuelHu, line 302: ?

src/RMGConvertLH5.cc
```diff
@@ -299,20 +299,28 @@ bool RMGConvertLH5::ConvertToLH5Internal() {
...
   for (auto& ntuple : ntuples) {
+    if (ntuple.size() == 0) LH5Log(RMGLog::fatal, "empty ntuple name, how is this possible?");
```
ManuelHu, line 304:
```suggestion
    if (ntuple.empty()) LH5Log(RMGLog::fatal, "empty ntuple name, how is this possible?");
```

otherwise `clang-tidy` will complain

gipert:
Very nice. One quick comment: let's rename fNtupleIDsS to fNtupleAuxIDs?

PR #426, opened 2025-07-04
Add `det<UID>` symlinks to the output and make the TCM use them

### LINKED ISSUE #421: TCM should be based on the detector UIDs?
At the moment, the TCM internally uses the index of the `tables` list stored in the HDF5 attribute of the TCM object itself. It would be more elegant to use the UID already used by remage internally. Unfortunately, `build_tcm()` is not very flexible in terms of hash functions, so there are few implementation options:

- add support for passing a mapping of table names -> uids to `build_tcm()`, then create this mapping in the Python wrapper
- if the the use-volume-names is activated in remage, add (hard?) links of the type `det<UID> -> <VOL_NAME>`. Then, we can use `build_tcm()` out of the box. These hard links would also make it very easy to use the TCM.

for the second option, I am not sure of where to make these hard links (`RMGConvertLH5` or wrapper). In both cases, the mapping UID <-> volume name must be made available, somehow.

gipert:
- [x] create HDF5 soft links to detector tables in a subgroup named `__by_uid__`
- [x] links are keyed as `det<REMAGE_UID>`, just like in the root stp group when volume names are not used
- [x] the TCM is now based on these UIDs
- [x] update documentation
- [x] check if everything works also in multithreaded mode with merging

Closes #421


docs/manual/output.md
```diff
@@ -392,19 +392,6 @@ output ntuples as a
 To use this feature, simply specify an output file with a `.lh5` extension, and
 _remage_ will perform the file conversion automatically.
 
-:::{note}
```
ManuelHu, line 395: you removed the opening tag of a note that is closed much later after the next paragraph... this should be restored

src/RMGConvertLH5.cc
```diff
@@ -292,19 +292,55 @@ bool RMGConvertLH5::ConvertToLH5Internal() {
...
+        }
+
+        // form soft link name "detUID" where UID is item.second.first.
+        auto soft_link_name = "uid" + std::to_string(item.second.first);
+        auto soft_link_name_full = ntuple_group_name + "/" + links_group_name + "/" + soft_link_name;
+        // do not create if the soft link already exists.
+        if (!ntuples_group.nameExists(soft_link_name)) {
+          // create a soft link to the current group itself.
+          herr_t err = H5Lcreate_soft(
+              ("/" + ntuple_group_name + "/" + ntuple).c_str(),
```
ManuelHu, line 331: there should be a c++ api for this, I guess? Such as `H5Location::link(...)`:

https://github.com/HDFGroup/hdf5/blob/2741b4079d07ec7016c6855a7e70a5e6af0099f4/c%2B%2B/src/H5Location.cpp#L1001-L1042

important note: you need to use the variant with two string arguments, other overloads create hard links...

wit this we can avoid the `c_str()`s and the explicit error handling

src/RMGConvertLH5.cc
```diff
@@ -292,19 +292,55 @@ bool RMGConvertLH5::ConvertToLH5Internal() {
...
+      if (item.second.second == ntuple) {
+        // create group named "links"
+        if (!ExistsByType(ntuples_group, links_group_name, H5O_TYPE_GROUP)) {
+          auto links_group = ntuples_group.createGroup(links_group_name);
+          links_group.close();
+        }
+
+        // form soft link name "detUID" where UID is item.second.first.
+        auto soft_link_name = "uid" + std::to_string(item.second.first);
+        auto soft_link_name_full = ntuple_group_name + "/" + links_group_name + "/" + soft_link_name;
```
ManuelHu, line 326: I am quite sure that clang-tidy will complain bitterly about this: https://clang.llvm.org/extra/clang-tidy/checks/performance/inefficient-string-concatenation.html

src/RMGConvertLH5.cc
gipert: I guess this is the best:

https://github.com/HDFGroup/hdf5/blob/2741b4079d07ec7016c6855a7e70a5e6af0099f4/c%2B%2B/src/H5Location.cpp#L1191

ManuelHu: certainly yes

gipert:
I think it's a good idea!

PR #433, opened 2025-07-08
Compile generic LLVM code in the container

### LINKED ISSUE #401: Numba cache with Docker containers
Starting remage takes long with Docker because of Numba compiling machine code. This code cannot be stored anywhere because the user in the container has no access to the host (e.g. home directory), so effectively the Numba caching functionality is lost. With Apptainer, things seem to work properly (I assume because the home directory is accessible).

This is quite annoying and we need to find a fix. We can't populate the cache at image build time since the machine code is architecture specific. Maybe we can pre-compile some more arch-independent code? Or maybe we can find a way to let Numba caching stuff somewhere at runtime?

gipert:
Closes #401.


gipert:
I tested it by building the container on sator and testing it on my laptop

PR #434, opened 2025-07-09
ci: split mt and non-mt test runs

ManuelHu:


ManuelHu:
> btw I don't think your approach will work, you can't do command expansion in CI afaik?

should work, it's just `sh`/`dash` not `bash`, I guess (so most _complex_ expansions do not work)? At least my `echo $(nproc)` just did its job

PR #435, opened 2025-07-09
gamma angular correlation validation

gipert:
Here's some work I did today on the topic, I won't have time to finish this for a while so if anyone wants to help I'd be happy :)

- [x] add gamma correlation tests / plots
- [x] think about how to boost statistics
- [x] add comparison with literature

Message from @jasondet:
> My code of course computes them on the fly from the nuclear polarity and the multipolarities assigned to each transition so that all gammas in a cascade can be properly correlated with each other, I'm attaching my summary of what the code actually does for your info (I sent it to the G4 guys a decade ago for a paper that's still "in progress"). The math can also produce a compact expression for the angular correlation between any two gammas. When I did validation, I calculated the coefficients myself using expressions in the attached paper from the 60's, see eq 3.46 and the text following it. Also useful for me was the attached tabulation by Taylor et al, and the attached Korean article from 2002 that summarizes the calculation for 2 consecutive gammas in modern notation with numerical values provided for 60Co. To perform the calcs, you might need the functions I implemented in [G4Clebsch](https://github.com/Geant4/geant4/blob/master/source/processes/hadronic/util/src/G4Clebsch.cc) and [G4PolarizationTransition](https://github.com/Geant4/geant4/blob/master/source/processes/hadronic/models/de_excitation/photon_evaporation/src/G4PolarizationTransition.cc).
>
> [G4GammaCorrDoc.pdf](https://github.com/user-attachments/files/21149626/G4GammaCorrDoc.pdf)
[1967RoseBrink_RevModPhys.39.306.pdf](https://github.com/user-attachments/files/21149628/1967RoseBrink_RevModPhys.39.306.pdf)
[1971Taylor_GGCorrCoeffs_NuclDataTablesA9_1.pdf](https://github.com/user-attachments/files/21149629/1971Taylor_GGCorrCoeffs_NuclDataTablesA9_1.pdf)
[2002Kim_JKorNuclSoc34_22.pdf](https://github.com/user-attachments/files/21149632/2002Kim_JKorNuclSoc34_22.pdf)

### General note

I think we should adopt this strategy to run tests: CTest only calls pytest, and all tests are written in python.

### DIFF
--- a/tests/gammacorr/CMakeLists.txt
+++ b/tests/gammacorr/CMakeLists.txt
@@ -11,8 +11,7 @@ endforeach()
 # run python tests using pytest
 add_test(NAME gammacorr/pytest-all COMMAND ${PYTHONPATH} -m pytest -s -vvv .)
 
-set_tests_properties(gammacorr/pytest-all PROPERTIES LABELS "mt;extra" ENVIRONMENT
-                                                     MPLCONFIGDIR=${CMAKE_SOURCE_DIR}/tests)
+set_tests_properties(gammacorr/pytest-all PROPERTIES LABELS "mt;extra;val;val-only")
 
 if(Geant4_VERSION VERSION_LESS "11.3.0")
   set_tests_properties(gammacorr/pytest-all PROPERTIES DISABLED True)


tests/hades/CMakeLists.txt
```diff
@@ -18,7 +21,8 @@ foreach(_mac ${_macros} ${_macros_extra} ${_macros_vis})
 endforeach()
 
 foreach(_mac ${_macros})
-  add_test(NAME hades-mt/${_mac} COMMAND ${REMAGE_PYEXE} -g gdml/main.gdml -t 2 macros/${_mac})
+  add_test(NAME hades-mt/${_mac} COMMAND ${REMAGE_PYEXE} -g gdml/main.gdml -t ${NUM_CORES}
+                                         macros/${_mac})
   set_tests_properties(hades-mt/${_mac} PROPERTIES LABELS mt)
```
ManuelHu, line 26: we could also set the PROCESSORS property...?

.github/workflows/main.yml
```diff
@@ -38,6 +38,12 @@ jobs:
           fetch-tags: true
           ref: ${{ github.event.pull_request && github.event.pull_request.head.sha || github.ref }}
 
+      - name: Get CPU core count
+        id: core_count
+        run: |
+          CORES=$(nproc --all)
```
ManuelHu, line 44: * shell expansion should work everywhere (at lerast for this simple case)
* I think `--all` is wrong. w/o it, nproc gives the number of usable cores (i.e., including system restrictions), which is what we should use. The number of hardware cores does not matter if the task affinity (or similar) disallows their use... (note: currently they are equal)

tests/gammacorr/test_angcorr.py
```diff
@@ -0,0 +1,105 @@
...
+            # read in dictionary with process ids, to filter later
+            _p = lh5.read("processes", remage_output)
+            processes = {k.decode(): int(v) for k, v in zip(_p.name, _p.procid)}
+
+            # group-by event id
+            trk = ak.unflatten(tracks, ak.run_lengths(tracks.evtid))
+
+            # select gamma particles created by a radioactive decay
+            gtrk = trk[
+                (trk.particle == 22) & (trk.procid == processes["RadioactiveDecay"])
```
ManuelHu, line 75: on G4 11.2 this is called `Radioactivation`

tests/gammacorr/test_angcorr.py
```diff
@@ -0,0 +1,105 @@
...
+            ]
```
ManuelHu, line 76:
```suggestion
            decay_procid = processes[
                (
                    "RadioactiveDecay"
                    if "RadioactiveDecay" in processes
                    else "Radioactivation"
                )
            ]
            gtrk = trk[(trk.particle == 22) & (trk.procid == decay_procid)]
```

ManuelHu:
we will also have to turn off this test for G4 11.2. I get quite a lot of segfaults with enabled gamma correlation (which is a known problem with G4 11.2 when you hit negative-parity states...)

ManuelHu:
no, 11.2 never got a fix for this bug (and, I guess, 11.3.0 also had it)

gipert:
> Then I think we should switch to requiring 11.3 (or was the fix in some 11.2 patch?)

though radioactive decays is only one application of remage. Maybe we can just add a warning to the docs, rather than requiring 11.3.whatever.

tests/gammacorr/test_angcorr.py
```diff
@@ -0,0 +1,139 @@
+from __future__ import annotations
+
+import awkward as ak
```
ManuelHu:
```suggestion
import os

import awkward as ak
```

PR #440, opened 2025-07-11
Fixes to output docs in user manual

EricMEsch:
Some small fixes i just found when reading through it.

### DIFF
--- a/docs/manual/output.md
+++ b/docs/manual/output.md
@@ -173,12 +173,14 @@ implement this functionality, for every event the total energy deposited is
 computed. This is based on summing the energy deposited in each `{UID}` added,
 or across all registered sensitive _Germanium_ detectors (if this macro command
 is not used). The event is then discarded if the energy is less than or equal to
-`ELOW` or less than `EHIGH`.
+`ELOW` or more than `EHIGH`.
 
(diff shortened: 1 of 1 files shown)


docs/manual/output.md
```diff
@@ -173,12 +173,14 @@ implement this functionality, for every event the total energy deposited is
...
-`ELOW` or less than `EHIGH`.
+`ELOW` or more than `EHIGH`.
 
 :::{note}
 
 This mechanism will remove the data from the event across all output schemes,
-not only the _Germanium_!
+not only the _Germanium_! However, `OutputSchemes` with their `StoreAlways()`
+function returning `true`, like the `TrackOutputScheme` or the
+`VertexOutputScheme` will always store their output.
```
ManuelHu, line 183:
```suggestion
function returning `true`, like the `RMGTrackOutputScheme` or the
`RMGVertexOutputScheme` will always store their output.
```

either we should use the full class names, or just use "the vertex or track output schemes" (i.e., not highlighted as code)

ManuelHu (commented):
apart from that, looks good

ManuelHu:
(also note that CI will fail)

PR #446, opened 2025-07-18
Fix an error occuring when setting the vertexgen late with bxdecay0

EricMEsch:
Right now if you use 

```macro
/RMG/Generator/Confine Volume
/RMG/Generator/Confinement/Physical/AddVolume germanium
```
After 
```
```
You will get an error, which is expected due to this line in `RMGGeneratorDecay0.cc`: 
```
if (!prim_gen) RMGLog::OutDev(RMGLog::fatal, "Primary position generator is nullptr")
```

this should be a fix for that and allow for any order of execution, as long as it is called before `/run/beamOn`.

@ManuelHu can you look over this and see if this is a solution we want? It feels kinda hacky, but i am not sure if there is a cleaner solution.

### DIFF
--- a/include/RMGGeneratorDecay0.hh
+++ b/include/RMGGeneratorDecay0.hh
@@ -21,6 +21,7 @@
 #include "G4GenericMessenger.hh"
 #include "G4ThreeVector.hh"
 
+#include "RMGMasterGenerator.hh"
 #include "RMGVGenerator.hh"
 #include "RMGVVertexGenerator.hh"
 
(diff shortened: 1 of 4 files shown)


ManuelHu:
for reference, see this bxdeay0 code:

https://github.com/BxCppDev/bxdecay0/blob/ee436dccd2958ff5f5c54df929f96dffbf90efd3/extensions/bxdecay0_g4/bxdecay0_g4/primary_generator_action.cc#L666-L688

EricMEsch:
Something like this? @ManuelHu

include/RMGGeneratorDecay0.hh
```diff
@@ -75,7 +76,7 @@ class RMGGeneratorDecay0 : public RMGVGenerator {
      *  @param prim_gen Pointer to the remage primary vertex generator.
```
ManuelHu, line 76: you also need to change this doc comment here

ManuelHu:
from my point of view, this is now ready and could be merged?

PR #447, opened 2025-07-21
initial changes for C++20

ManuelHu:
* switch standard and fix the first compile errors
* wrap fmt format strings in `fmt::runtime()` calls
* switch in some places to `std::format`. Unfortunately, it only supports constexpr strings (there is no equivalent to `fmt::runtime` in stdlib in C++20)
* disable the compile-time warning of non-matching ROOT C++ standards.
* fix new clang-tidy warnings


gipert:
Should we merge this for now?

src/CMakeLists.txt
```diff
@@ -121,6 +121,9 @@ endif()
 # Ensure clients are aware of the minimum C++ standard we were compiled with
 target_compile_features(remage PUBLIC cxx_std_${CMAKE_CXX_STANDARD})
 
+# for initial C++20 compatibility with a C++17 compiled ROOT image. TODO: remove later
```
gipert, line 124: we should update the remage-base image and compile everything with C++20...

PR #454, opened 2025-07-30
Enable gamma angular correlations by default

gipert:
- **phy: rename ThermalScattering to EnableNeutronThermalScattering**
- **phy: enable nuclear gamma angular correlations by default**


gipert:
uhm correlations are still disabled...

PR #457, opened 2025-08-07
Add unit tests and validation plots for bxdecay

EricMEsch:
This currently runs two tests which at the same time produce output for validation plots.

These are the energy of the two primary electrons (the neutrinos are not even created with bxdecay)
e-combined-primary-energy output

And the opening angle between the primary electrons.
e-primary-opening-angle output

I would say the energy looks good, i am not sure about the opening angles. I have to admit i have no idea what we would expect there. 

The main issue is that we need high statistics for good plots, and that means longer simulation times.

I still need to add the plots to the validation suite, so this is just a draft, but i guess you can already comment on it.

Also i had to include the `/RMG/Geometry/RegisterDetector Germanium det1 001` in the macros because otherwise the `/RMG/Output/Vertex/StorePrimaryParticleInformation` command would not exist. I assume this is because the simulation would run in non-persistency mode. But this means right now we write away information we do not need in the tests i added


tests/bxdecay/run_sims.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+
+from remage import remage_run
+
+bbv0_commands = [
+    "macros/template.mac",
+    "/RMG/Generator/BxDecay0/DoubleBetaDecay Ge76 0vbb",
+    "/run/beamOn 10000",
```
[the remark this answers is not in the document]
gipert: yes I couldn't think of a better way

ManuelHu:
> I guess i also have to add some conditional code to the CMakeLists to only include the test if remage is built with bxdecay support

yes. we alredy have something similar for some other tests here: https://github.com/legend-exp/remage/blob/f5b726f90b5c56c5b6b1b41db8ca0784bc72d988/tests/hades/CMakeLists.txt#L29-L32

gipert:
Also see https://remage.readthedocs.io/en/stable/developer.html#configuring-cmake for CMake tips related to plots for the validation report.

EricMEsch:
I think it would be pretty useful, but i can not think of a simple solution... You can either
- Skip the entire Geant4 initialization which might cause many problems
- Intercept the particle before the `GeneratePrimaries()` function? But i doubt Geant4 would be happy with that
- Try to immediately discard the primary track on the first step including all secondaries maybe produced... But that seems like an inefficient workaround

gipert:
>Try to immediately discard the primary track on the first step including all secondaries maybe produced... But that seems like an inefficient workaround

why do you think this would be inefficient? I think this is the best solution

EricMEsch:
Well Geant4 has to set up the entire world, physics databases etc. anyways. I am not really talking about a quantitative inefficiency, which might only be a few %. I more meant a logical/qualitative inefficiency.

As you are initializing the entire Framework and then only access a small isolated part. From a logical point of view the "cleanest" way would be to just access the isolated part on its own... But that would be rather difficult to implement i think

gipert:
I think we should try to simulate a lot, for smooth histograms. Much more than what's in your tests here. In such a case, the app initialization time would be much lower than initialization time.

gipert:
yes, could you quickly test it?

gipert:
can you post again the plots? how's the stats now?

tdixon97:
You could reduce the binning, this will mean the CPU demands are less

gipert:
go back on holiday smart ass

tests/bxdecay/plot_decay_modes.py
```diff
@@ -0,0 +1,99 @@
...
+
+def plot_angles():
+    fig, ax = plt.subplots()
+
+    angles_0vbb = get_primary_electron_angle("0vbb.lh5")
+    angles_2vbb = get_primary_electron_angle("2vbb.lh5")
+    angles_0vbb_M1 = get_primary_electron_angle("0vbb_M1.lh5")
+    angles_0vbb_lambda_0 = get_primary_electron_angle("0vbb_lambda_0.lh5")
+
+    modes = ["0vbb", "2vbb", "0vbb_M1", "0vbb_lambda_0"]
```
gipert, line 79: can we use prettier legend labels? I mean using greek letters (vbb -> νββ, lambda -> λ, ...)

docs/validation/bxdecay.md
```diff
@@ -0,0 +1,36 @@
...
+In this section we take a look at the primary spectrum produced by the
+_bxdecay0_ package. The main documentation for this extension is available
+[here](https://github.com/BxCppDev/bxdecay0). It aims to reproduce reasonable
+primary particles to be used as generator for double beta decay physics.
+
+## Primary electron properties
+
+The main thing we can compare here are the properties of the primary electrons
+produced. The _bxdecay0_ package actually does not even produce the neutrinos
+emitted in e.g. `2vbb`.
```
gipert, line 12: see comment about decay mode names

docs/validation/index.md
```diff
@@ -13,6 +13,7 @@ versioninfo
 :caption: Sections
 :maxdepth: 2
 
+bxdecay
```
gipert, line 16:
```suggestion
bxdecay0
```
also check correct spelling everywhere else please

docs/validation/bxdecay.md
```diff
@@ -0,0 +1,36 @@
+# Checking bxdecay0
```
gipert, line 1:
```suggestion
# Double-beta decay physics
```

src/RMGSteppingAction.cc
```diff
@@ -92,6 +94,14 @@ void RMGSteppingAction::DefineCommands() {
       .SetParameterName("max_lifetime", false)
       .SetDefaultValue("-1")
       .SetStates(G4State_Idle);
+
+  fMessenger->DeclareProperty("SkipTracking", fSkipTracking)
+      .SetGuidance(
+          "Immediately discard any tracks after primary generation. To verify primary generator."
```
gipert, line 100:
```suggestion
          "Immediately discard tracks after primary particle generation. This feature is meant for debugging primary generation."
```

gipert (changes requested):

tests/bxdecay/plot_decay_modes.py
```diff
@@ -0,0 +1,99 @@
...
+        ).Double()
+        h.fill(angle * 180 / np.pi)
+        h.plot(ax=ax, yerr=False, label=f"{decay_mode}")
+
+    ax.set_xlabel("Angle between primary electrons [°]")
+    ax.set_ylabel("Density")
+    ax.legend()
+    ax.grid()
+
+    fig.savefig("e-primary-opening-angle.output.png")
```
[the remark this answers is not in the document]
EricMEsch: But they are under the folder bxdecay? They always are under the bxdecay folder or have the prefix `bxdecay/` anywhere in the tests, validation suite and wherever they get moved i am fairly certain.

tests/bxdecay/plot_decay_modes.py
EricMEsch: But i would not use them here? Here it just sets the (for the tests internal) output file names and i would not name the output files with greek letters...  also i use the same name as for the macro command and i dont think having the options of macro commands being greek letters is a good idea

EricMEsch:
oke, everything fixed now

PR #461, opened 2025-08-20
Add `GeneralDetector` and `GeneralOutputScheme`

EricMEsch:


EricMEsch:
I am also not sure, what other things should maybe adjusted for this... Especially in the post-processing

tdixon97:
Yeah maybe, but the idea is to be able to still keep the tracking info, just with larger clusters. I was hoping that the code structure would avoid any code duplication

gipert:
So you still want to look at where the energy is deposited?

tdixon97:
I guess it depends the application? Eg. in LEGEND sims maybe not, but imagine you had a detector with a very large (inactive shield), eg lead maybe there its useful?

gipert:
We can keep the approach of this PR, but we should think of whether there is a better way to structure the code to avoid duplication. One straightforward idea is to create yet another new class inheritance layer (ugly), or move the methods to the base output class.

tdixon97:
I think that methods should be moved as much as possible to the RMGOutputTools namespace

gipert:
I'm not sure, we would still have duplicated headers. The least duplication would be achieved with class inheritance.

tdixon97:
What I see duplicated in this case is mainly "ProcessHits" maybe that can be refactored a bit to use some generic methods?

EricMEsch:
But i agree with luigi, that just a dumb Calorimeter would also be something that is pretty useful. Minimal output for larger volumes where you still want the total energy deposited...

Although a time-coincidence would still be pretty useful for that

gipert:
I understand that this general scheme might be useful, but maybe we should just implement what we need for now. This `Calorimeter` scheme is pretty simple to implement and does what we need.

EricMEsch:
That would be enough for my use case. Calorimeter with a time-grouping and timestamps. I am not sure that is enough for Tobys/Rosannas use case?

EricMEsch:
Well if for my cosmogenic simulation there is energy deposition in a volume i would like to know when. 

As i have to differentiate between prompt and delayed energy deposition. So i would like to summarize the depositions into hits on the time-scale and give them a timestamp.

So not only energy deposited per event, but e.g. summarize all energy depositions within 10 us together with one timestamp.

So the output would be an energy vector and a timestamp vector. Still very reduced (and also have the option to set the "timewindow" per macro command so you can just set it to a big number if you want to). 

But just energy deposited per event is not very useful for cosmogenics if you can not differentiate between prompt and delayed

tdixon97:
I think you are saying you still want the time-windowing (effectively)? Actually I did not think much about what happens with the time condition in the pre-clustering

gipert:
Yes @EricMEsch honestly this sounds like a specialized scheme, not something that should be supported by a generic scheme.

PR #462, opened 2025-08-21
Increase tolerance in checks of surface sampling

tdixon97:


tests/confinement/test_basic_surface.py
```diff
@@ -35,7 +35,7 @@ def add_local_pos(vertices, pos):
     return vertices
 
 
-tol = 1e-6  # mm
+tol = 1e-5  # mm
```
tdixon97:
```suggestion
tol = 1e-4  # mm
```

tdixon97:
@ManuelHu I increased alot the threshold here -- hopefully it fixes the problem, but really someone should look into this in more detail.

ManuelHu:
https://github.com/tdixon97/remage/blob/37644db796633306737e0d877a26d45c422c6750/tests/confinement/test_basic_surface.py#L42-L46

am I stupid or are we not taking into account the `tol` for some of these comparsions, i.e. `(np.sqrt(x**2 + y**2) < 60)`?? Similar for the "middle comparisons" in the boxes case.

And `abs(z) < 30 - tol` is wrong, it should be `abs(z-30)<tol`, right? similar for others

tdixon97:
Closing since its superseded by #670

Issue #450, opened 2025-07-29
Standardize Tests


EricMEsch:
Luigi recently proposed this idea in a PR, to have a general-purpose geometry somewhere that is used by all (most) non-geometry-critical tests.

EricMEsch:
But its probably reasonable to have a set of geometries. As most of the tests are not geometry-critical, but some require an optical volume etc. while others only need germanium in a world volume. Others then require at least two germaniums... So in order to have fast tests, you would go to the geometry which satisfies your minimum requirements.

Or would you rather opt for one more complex geometry that is used by most of the tests?

ManuelHu:
> like the replace_lines() function and similar code

it would be interesting if we actually need most of those helpers. I guess, with the recent additions of loading commands from the CLI args and the alias feature, we should be able to use run_remage without temporary macro files in most cases?

example: in your PR #455 this would be two args; first pass the register dets command and then the macro file name.

I think we can remove quite some test code from times where all this was not yet possible.

EricMEsch:
I was not sure if this would be working if you need to load multiple commands from CLI? I tried and it did not like the `\n` command. There probably is another way of passing on multiple macro commands but i didnt figure it out

ManuelHu:
Just passing multiple space-seperated strings like `"command 1" "command 2" "filename"` should work.

Am 4. August 2025 17:05:43 MESZ schrieb Eric Esch ***@***.***>:
>EricMEsch left a comment (legend-exp/remage#450)
>
>I was not sure if this would be working if you need to load multiple commands from CLI? I tried and it did not like the `\n` command. There probably is another way of passing on multiple macro commands but i didnt figure it out
>

ManuelHu:
I meant passing as plain arguments, not as alias. Commands cannot be used as aliases (only macro parameter values can)

Also you need to use `--` to properly terminate the püarsing of the substitution argument...

`remage -s TEST=Volume -- "/RMG/Generator/Confine {TEST}" "/run/initialize"` works for me

EricMEsch:
Okay, but how would i properly use this with `remage_run`? As part of the `macros` argument i would assume? 
I would first pass the `template.mac` and then have to call `/run/beamOn` myself later as part of my arguments and not in `template.mac`? Or would you opt to generally avoid the `.mac` file and pass everything myself?

PR #464, opened 2025-08-22
Enable the self-hosted runner

### LINKED ISSUE #423: Setup our own CI runner
We need runners with more core in order to make development more efficient and have a meaningful validation suite. We have an unused 32-core server at TUM that we could set up in ~1 day of work.

gipert:
32-core / 128 GB RAM server at TUM.

Closes #423.


gipert:
no tobi it's just a joke

gipert:
vtw i made a mistake in the config, needs to be configured as a matrix. i will change it later

tdixon97:
Seems great to me, first step towards a proper and serious validation suite

gipert:
I think that locally one would want to usually only run unit tests, which are never cpu intensive. in case a piece of the validation suite is being developed, then usually one would run only that specific test. so I think it's fine, I would not now support producing a low-stats version of the validation.

We should handle all this with test labels, like we do already.

This is also related to #450, as a general effort towards getting a more coherently structured test suite.

gipert:
I would merge this for now, we can do the rest of the work in a separate PR.

PR #466, opened 2025-08-31
add regex support for copy_nr in RegisterDetector

ManuelHu:


src/RMGHardwareMessenger.cc
```diff
@@ -91,14 +91,13 @@ void RMGHardwareMessenger::RegisterDetectorCmd(const std::string& parameters) {
   auto type = RMGTools::ToEnum<RMGDetectorType>(std::string(type_str), "detector type");
   auto pv_name = next();
   const int uid = std::stoi(next());
-  int copy_nr = 0;
   auto copy_nr_str = next();
-  if (!copy_nr_str.empty()) copy_nr = std::stoi(copy_nr_str);
+  if (!copy_nr_str.empty()) copy_nr_str = "0";
```
EricMEsch, line 95: I am not sure if i am misunderstanding this. But if the string is **not** empty (which means the user specified something) you reset it to the default value `"0"`? Am i missing something?

PR #475, opened 2025-09-03
Ib update

311Francis:
The new Inner Bremsstrahlung update after adding the CMake file with some changes. I started a new branch since the original one was broken for some reason. The pulled files are different from what is shown on github.

### DIFF
--- a/include/RMGInnerBremsstrahlungProcess.hh
+++ b/include/RMGInnerBremsstrahlungProcess.hh
@@ -1,0 +1,126 @@
+#ifndef RMG_INNER_BREMSSTRAHLUNG_PROCESS_HH
+#define RMG_INNER_BREMSSTRAHLUNG_PROCESS_HH
+
+#include "G4ParticleDefinition.hh"
+#include "G4Step.hh"
+#include "G4Track.hh"
+#include "G4VParticleChange.hh"
+#include "G4WrapperProcess.hh"
+#include "globals.hh"
+#include "G4Positron.hh"
+
+class RMGInnerBremsstrahlungProcess : public G4WrapperProcess {
+
+  public:
+
+    /**
+     * @brief Constructs a new RMG Inner Bremsstrahlung wrapper process.
+     *
+     * @param aNamePrefix Prefix for naming the process (default "RMG_IB").
+     * @param aType Process type (default @c fDecay).
+     */
+    explicit RMGInnerBremsstrahlungProcess(const G4String& aNamePrefix = "RMG_IB",
+                                          G4ProcessType aType = fDecay);
+
+    /**
+     * @brief Virtual destructor.
+     */
+    virtual ~RMGInnerBremsstrahlungProcess() = default;
+
+    /**
+     * @brief Applies Inner Bremsstrahlung generation after radioactive decay.
+     *
+     * This method overrides @c PostStepDoIt() of @ref G4WrapperProcess. It first calls the
+     * wrapped decay process's @c PostStepDoIt() to obtain the decay products. Then it examines
+     * all secondary particles, identifies beta electrons, calculates the Inner Bremsstrahlung
+     * probability based on the electron energy, and generates IB photons accordingly. The IB
+     * photons are created with proper kinematic properties and added to the secondary stack.
+     *
+     * @param aTrack The current track undergoing decay.
+     * @param aStep The current step.
+     * @return Pointer to the particle change with added IB photons.
+     */
+    G4VParticleChange* PostStepDoIt(const G4Track& aTrack, const G4Step& aStep) override;
+
+    /**
+     * @brief Enables or disables Inner Bremsstrahlung generation.
+     *
+     * @param enabled True to enable IB generation, false to disable.
+     */
+    void SetEnabled(G4bool enabled) { fEnabled = enabled; }
+
+    /**
+     * @brief Checks if Inner Bremsstrahlung generation is enabled.
+     *
+     * @return True if enabled, false otherwise.
+     */
+    G4bool IsEnabled() const { return fEnabled; }
+
+    /**
+     * @brief Sets a scaling factor for IB probability (for systematic studies).
+     *
+     * @param scale Probability scaling factor (default 1.0).
+     */
+    void SetIBProbabilityScale(G4double scale) { fIBProbabilityScale = scale; }
+
+    /**
+     * @brief Gets the current IB probability scaling factor.
+     *
+     * @return Probability scaling factor.
+     */
+    G4double GetIBProbabilityScale() const { return fIBProbabilityScale; }
+
+  private:
+
+    /**
+     * @brief Calculates the phi function for Inner Bremsstrahlung spectrum.
+     *
+     * @param W_prime Dimensionless electron energy before photon emission.
+     * @param omega Dimensionless photon energy.
+     * @return Value of the phi function.
+     */
+    G4double PhiFunction(G4double W_prime, G4double omega);
+
+    /**
+     * @brief Calculates the total Inner Bremsstrahlung probability for a given electron energy.
+     *
+     * @param electronEnergy Kinetic energy of the beta electron.
+     * @return Total IB probability (0 to 1).
+     */
+    G4double CalculateIBProbability(G4double electronEnergy);
+
+    /**
+     * @brief Samples a photon energy from the Inner Bremsstrahlung spectrum.
+     *
+     * @param electronEnergy Kinetic energy of the beta electron.
+     * @return Sampled IB photon energy.
+     */
+    G4double SamplePhotonEnergy(G4double electronEnergy);
+
+    /**
+     * @brief Generates IB photons for beta electrons in the decay secondaries.
+     *
+     * @param particleChange Particle change object containing decay secondaries.
+     * @param parentTrack The original decaying nucleus track.
+     * @param aStep The current step.
+     */
+    void GenerateInnerBremsstrahlungForSecondaries(G4VParticleChange* particleChange,
+                                                  const G4Track& parentTrack,
+                                                  const G4Step& aStep);
+
+    /**
+     * @brief Checks if a secondary track is a beta electron from decay.
+     *
+     * @param track Secondary track to check.
+     * @return True if it's a beta electron, false otherwise.
+     */
+    G4bool IsBetaElectron(G4Track* track);
+
+    // Physical constants
+    G4double fAlpha;                    ///< Fine structure constant (1/137)
+    G4double fElectronMass;
+    G4bool fEnabled;
+    G4double fIBProbabilityScale;
+};
+
+#endif
--- a/include/RMGInnerBremsstrahlungProcess.hh
+++ b/include/RMGInnerBremsstrahlungProcess.hh
@@ -2,12 +2,12 @@
 #define RMG_INNER_BREMSSTRAHLUNG_PROCESS_HH
 
 #include "G4ParticleDefinition.hh"
+#include "G4Positron.hh"
 #include "G4Step.hh"
 #include "G4Track.hh"
 #include "G4VParticleChange.hh"
 #include "G4WrapperProcess.hh"
 #include "globals.hh"
-#include "G4Positron.hh"
 
 class RMGInnerBremsstrahlungProcess : public G4WrapperProcess {
 
@@ -19,8 +19,10 @@ class RMGInnerBremsstrahlungProcess : public G4WrapperProcess {
      * @param aNamePrefix Prefix for naming the process (default "RMG_IB").
      * @param aType Process type (default @c fDecay).
      */
-    explicit RMGInnerBremsstrahlungProcess(const G4String& aNamePrefix = "RMG_IB",
-                                          G4ProcessType aType = fDecay);
+    explicit RMGInnerBremsstrahlungProcess(
+        const G4String& aNamePrefix = "RMG_IB",
+        G4ProcessType aType = fDecay
+    );
 
     /**
      * @brief Virtual destructor.
@@ -104,9 +106,11 @@ class RMGInnerBremsstrahlungProcess : public G4WrapperProcess {
      * @param parentTrack The original decaying nucleus track.
      * @param aStep The current step.
      */
-    void GenerateInnerBremsstrahlungForSecondaries(G4VParticleChange* particleChange,
-                                                  const G4Track& parentTrack,
-                                                  const G4Step& aStep);
+    void GenerateInnerBremsstrahlungForSecondaries(
+        G4VParticleChange* particleChange,
+        const G4Track& parentTrack,
+        const G4Step& aStep
+    );
 
     /**
      * @brief Checks if a secondary track is a beta electron from decay.
@@ -117,7 +121,7 @@ class RMGInnerBremsstrahlungProcess : public G4WrapperProcess {
     G4bool IsBetaElectron(G4Track* track);
 
     // Physical constants
-    G4double fAlpha;                    ///< Fine structure constant (1/137)
+    G4double fAlpha; ///< Fine structure constant (1/137)
     G4double fElectronMass;
     G4bool fEnabled;
     G4double fIBProbabilityScale;
--- a/include/RMGInnerBremsstrahlungProcess.hh
+++ b/include/RMGInnerBremsstrahlungProcess.hh
@@ -121,8 +121,6 @@ class RMGInnerBremsstrahlungProcess : public G4WrapperProcess {
     G4bool IsBetaElectron(G4Track* track);
 
     // Physical constants
-    G4double fAlpha; ///< Fine structure constant (1/137)
-    G4double fElectronMass;
     G4bool fEnabled;
     G4double fIBProbabilityScale;
 };
--- a/include/RMGPhysics.hh
+++ b/include/RMGPhysics.hh
@@ -137,7 +137,7 @@ class RMGPhysics : public G4VModularPhysicsList {
     virtual void ConstructOptical();
 
   private:
-    
+
     G4bool fUseInnerBremsstrahlung = false;
     ProdCutStore fProdCuts = {};
     ProdCutStore fProdCutsSensitive = {};
--- a/include/RMGPhysics.hh
+++ b/include/RMGPhysics.hh
@@ -137,7 +137,8 @@ class RMGPhysics : public G4VModularPhysicsList {
     virtual void ConstructOptical();
 
   private:
-
+    
+    G4bool fUseInnerBremsstrahlung = false;
     ProdCutStore fProdCuts = {};
     ProdCutStore fProdCutsSensitive = {};
     bool fConstructOptical = false;
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,6 +18,7 @@ set(PROJECT_PUBLIC_HEADERS
     ${_root}/include/RMGGermaniumDetector.hh
     ${_root}/include/RMGGermaniumOutputScheme.hh
     ${_root}/include/RMGGrabmayrGCReader.hh
+    ${_root}/include/RMGInnerBremsstrahlungProcess.hh
     ${_root}/include/RMGIsotopeFilterScheme.hh
     ${_root}/include/RMGIpc.hh
     ${_root}/include/RMGLog.hh
@@ -65,6 +66,7 @@ set(PROJECT_SOURCES
     ${_root}/src/RMGGermaniumDetector.cc
     ${_root}/src/RMGGermaniumOutputScheme.cc
     ${_root}/src/RMGGrabmayrGCReader.cc
+    ${_root}/include/RMGInnerBremsstrahlungProcess.cc
     ${_root}/src/RMGIpc.cc
     ${_root}/src/RMGIsotopeFilterScheme.cc
     ${_root}/src/RMGLog.cc
@@ -120,9 +122,6 @@ endif()
 
 # Ensure clients are aware of the minimum C++ standard we were compiled with
 target_compile_features(remage PUBLIC cxx_std_${CMAKE_CXX_STANDARD})
-
-# for initial C++20 compatibility with a C++17 compiled ROOT image. TODO: remove later
-target_compile_options(remage PRIVATE -Wno-cpp)
 
 target_include_directories(remage PUBLIC "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>"
                                          $<INSTALL_INTERFACE:include/remage>)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -66,7 +66,7 @@ set(PROJECT_SOURCES
     ${_root}/src/RMGGermaniumDetector.cc
     ${_root}/src/RMGGermaniumOutputScheme.cc
     ${_root}/src/RMGGrabmayrGCReader.cc
-    ${_root}/include/RMGInnerBremsstrahlungProcess.cc
+    ${_root}/src/RMGInnerBremsstrahlungProcess.cc
     ${_root}/src/RMGIpc.cc
     ${_root}/src/RMGIsotopeFilterScheme.cc
     ${_root}/src/RMGLog.cc
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -123,6 +123,8 @@ endif()
 # Ensure clients are aware of the minimum C++ standard we were compiled with
 target_compile_features(remage PUBLIC cxx_std_${CMAKE_CXX_STANDARD})
 
+# for initial C++20 compatibility with a C++17 compiled ROOT image. TODO: remove later
+target_compile_options(remage PRIVATE -Wno-cpp)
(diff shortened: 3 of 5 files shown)


src/RMGInnerBremsstrahlungProcess.cc
```diff
@@ -0,0 +1,255 @@
...
+    omegas.push_back(omega);
+    sum += value;
+    cdf.push_back(sum);
+  }
+
+  // Normalize the CDF
+  if (sum > 0.0) {
+    for (size_t i = 0; i < cdf.size(); i++) { cdf[i] /= sum; }
+  } else {
+    return 0.01 * fElectronMass; // Fallback
```
ManuelHu, line 241:
```suggestion
    return 0.01 * CLHEP::electron_mass_c2; // Fallback
```

src/RMGInnerBremsstrahlungProcess.cc
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  // Sample from the CDF
+  G4double r = G4UniformRand();
+  for (size_t i = 0; i < cdf.size(); i++) {
+    if (r <= cdf[i]) {
+      // Convert back to energy in keV
+      return omegas[i] * fElectronMass;
```
ManuelHu, line 249:
```suggestion
      return omegas[i] * CLHEP::electron_mass_c2;
```

src/RMGInnerBremsstrahlungProcess.cc
```diff
@@ -0,0 +1,255 @@
...
+    }
+  }
+
+  // Fallback to the last value
+  return omegas.back() * fElectronMass;
```
ManuelHu, line 254:
```suggestion
  return omegas.back() * CLHEP::electron_mass_c2;
```

include/RMGInnerBremsstrahlungProcess.hh
```diff
@@ -0,0 +1,130 @@
...
+     * @brief Checks if a secondary track is a beta electron from decay.
+     *
+     * @param track Secondary track to check.
+     * @return True if it's a beta electron, false otherwise.
+     */
+    G4bool IsBetaElectron(G4Track* track);
+
+    // Physical constants
+    G4double fAlpha; ///< Fine structure constant (1/137)
+    G4double fElectronMass;
```
ManuelHu, line 125:
```suggestion
```

src/RMGInnerBremsstrahlungProcess.cc
```diff
@@ -0,0 +1,255 @@
...
+      if (electronEnergy < 0.0 * CLHEP::keV) { continue; }
+
+      // Calculate IB probability (with optional scaling)
+      G4double ibProbability = CalculateIBProbability(electronEnergy) * fIBProbabilityScale;
+
+      RMGLog::OutFormat(
+          RMGLog::debug,
+          "{}: Beta electron energy: {:.3f} keV, IB probability: {:.6f}",
+          GetProcessName(),
+          electronEnergy / keV,
```
ManuelHu, line 103:
```suggestion
          electronEnergy / CLHEP::keV,
```

src/RMGInnerBremsstrahlungProcess.cc
```diff
@@ -0,0 +1,255 @@
...
+                                                 ->GetTrackingManager()
+                                                 ->GetSteppingManager();
+        steppingManager->GetfSecondary()->push_back(ibGammaTrack);
+
+        RMGLog::OutFormat(
+            RMGLog::debug,
+            "{}: Generated IB photon {:.3f} keV from beta {:.3f} keV",
+            GetProcessName(),
+            gammaEnergy / keV,
+            electronEnergy / keV
```
ManuelHu, line 149:
```suggestion
            gammaEnergy / CLHEP::keV,
            electronEnergy / CLHEP::keV
```

src/RMGInnerBremsstrahlungProcess.cc
```diff
@@ -0,0 +1,253 @@
...
+
+#include "G4DynamicParticle.hh"
+#include "G4Electron.hh"
+#include "G4EventManager.hh"
+#include "G4Gamma.hh"
+#include "G4PhysicalConstants.hh"
+#include "G4Positron.hh"
+#include "G4Step.hh"
+#include "G4SteppingManager.hh"
+#include "G4SystemOfUnits.hh"
```
ManuelHu, line 15:
```suggestion
```

src/RMGInnerBremsstrahlungProcess.cc
```diff
@@ -0,0 +1,253 @@
...
+  // Generate Inner Bremsstrahlung for any beta electrons in the secondaries
+  GenerateInnerBremsstrahlungForSecondaries(particleChange, aTrack, aStep);
+
+  return particleChange;
+}
+
+void RMGInnerBremsstrahlungProcess::GenerateInnerBremsstrahlungForSecondaries(
+    G4VParticleChange* particleChange,
+    const G4Track& parentTrack,
+    const G4Step& aStep
```
ManuelHu, line 63:
```suggestion
    const G4Step&
```

src/RMGInnerBremsstrahlungProcess.cc
```diff
@@ -0,0 +1,253 @@
...
+) {
+
+  // First check if this decay actually produces electrons/positrons (beta decay)
+  bool hasBetaParticles = false;
+  for (G4int i = 0; i < particleChange->GetNumberOfSecondaries(); i++) {
+    G4Track* track = particleChange->GetSecondary(i);
+    if (track->GetDefinition() == G4Electron::Definition() ||
+        track->GetDefinition() == G4Positron::Definition()) {
```
ManuelHu, line 71:
```suggestion
    if (IsBetaElectron(track)) {
```

src/CMakeLists.txt
```diff
@@ -123,7 +125,6 @@ target_compile_features(remage PUBLIC cxx_std_${CMAKE_CXX_STANDARD})
 
 # for initial C++20 compatibility with a C++17 compiled ROOT image. TODO: remove later
 target_compile_options(remage PRIVATE -Wno-cpp)
```
ManuelHu, line 127:
```suggestion
target_compile_options(remage PRIVATE -Wno-cpp)

```

src/RMGPhysics.cc
```diff
@@ -305,6 +306,44 @@ void RMGPhysics::ConstructProcess() {
...
+
+  // Add Inner Bremsstrahlung wrapper for radioactive decays
+  if (fUseInnerBremsstrahlung) {
+    RMGLog::Out(RMGLog::detail, "Adding Inner Bremsstrahlung physics");
+
+    // Get all ions from the particle table
+    auto particleIterator = GetParticleIterator();
+    particleIterator->reset();
+    while ((*particleIterator())()) {
+      auto particle = particleIterator->value();
```
ManuelHu, line 318:
```suggestion
    auto particleIterator = GetParticleIterator();
    particleIterator->reset();
    while ((*particleIterator)()) {
      auto particle = particleIterator->value();
```

ManuelHu:
@311Francis I have allowed myself to implement some of the changes from the last PR, and to fix some other comments from Toby and me. I hope that this is okay with you :-) Now it  compiles on all of our test targets (macOS & linux), so we can proceed with functional testing.

311Francis:
> Hi Francis, In general we divide testing into two parts:
> 
> ### Unit tests
> We want to have tests on the basic methods to see that they do what we expect and handle edge cases etc. In python we do this with `pytest` (see [link](https://learn.scientific-python.org/development/guides/pytest/)). In C++ its a bit harder to write tests instead we use Ctest (which is not as good). Here is a bit of info [[link]](https://remage.readthedocs.io/en/stable/developer.html#testing). Generally you have to write a little c++ script that links to your module and runs the functions checking the inputs are what we expect.
> 
> For example you can have a look at what I did for the surface sampling methods [[here] ](https://github.com/legend-exp/remage/blob/main/tests/confinement/test-surface-sampler-methods.cpp).
> 
> This is well suited to the functions in `RMGInnerBremstralhungProcess` such as the `PhiFunction` etc. This will also tell you if your code is not so well structured since in that case writing a test for a function becomes hard (usually if a function is trying to do too many different things).
> 
> ### Validation
> Independently from that since testing every part of a Geant4 simulation is very hard, we typically do some "validation", this means we run a full simulations and we look at the results. This is implemented in a similar way but the output is usually just some plots that get added to validation report [[link]](https://legend-exp.github.io/remage/validation/v0.13.1/).
> 
> Essentially this means we write a remage macro and GDML files to run some simulations and then we also write a python script to make plots or checks of the output. You can start doing all of this outside of the testing enviroment and when its ready we can add it.
> 
> I think then we should add a new section called "Physics Processes" which would have some plots showing what physics processes implemented in remage are doing.
> 
> We need to think of a very simple way to test the essential functionality of the code, for example:
> 
> * Generate electrons from a simple beta decay (eg 39Ar) and compare the energy distribution of the e- and IB photons to the e- with IB turned off
> * Generate electrons from Ar39 in LAr and compare the energy deposited in Ge (maybe with some post-processing in reboost).
> 
> Does this all make sense?

Got it, I will work on it. For the unit testing, I can just test it locally on my remage repo or do you recommend me to make a copy and test it there.

tdixon97:
I am not sure I understand what you mean? Why don't you start with a script locally to develop the tests and then when its ready add it to the "test" directory.

311Francis:
Like if I want to test it, I need to create the build folder for testing (ctest). I was wondering should I do it in my local repo, since if I try to create a commit or update it in the future, the build folder that I create for testing might also be pushed on github.

tdixon97:
For now just do something seperate from your remage install (but ofc course linking remage). When we add it to the tests everything gets handled by ctest so you dont need to worry

311Francis:
Sorry for the late update. Dr. Henning and I were working on models for Inner Brem generated from electron capture and double beta decay. BUt just for the current model of pure beta plus and beta minue, I was working on the unit test and the function. I made an adjustment on RMGTrackOutputScheme.CC. I changed the

`fTrackEntries.emplace(
        G4EventManager::GetEventManager()->GetConstCurrentEvent()->GetEventID(),
        track->GetTrackID(),
        track->GetParentID(),
        proc_id,
        primary->GetPDGcode(),
        track->GetGlobalTime(),
        pos.getX(),
        pos.getY(),
        pos.getZ(),
        primary->GetMomentum().getX(),
        primary->GetMomentum().getY(),
        primary->GetMomentum().getZ(),
        track->GetKineticEnergy()
    );`
to 
`fTrackEntries.push_back({
        G4EventManager::GetEventManager()->GetConstCurrentEvent()->GetEventID(),
        track->GetTrackID(),
        track->GetParentID(),
        proc_id,
        primary->GetPDGcode(),
        track->GetGlobalTime(),
        pos.getX(),
        pos.getY(),
        pos.getZ(),
        primary->GetMomentum().getX(),
        primary->GetMomentum().getY(),
        primary->GetMomentum().getZ(),
        track->GetKineticEnergy()
    });` due to compile error below: `In file included from /Users/zichenwang/remage/src/RMGTrackOutputScheme.cc:16:
In file included from /Users/zichenwang/remage/include/RMGTrackOutputScheme.hh:19:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/map:588:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/__memory/allocator.h:15:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/__memory/allocate_at_least.h:13:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/__memory/allocator_traits.h:304:9: error: no matching function for call to 'construct_at'
        _VSTD::construct_at(__p, _VSTD::forward<_Args>(__args)...);
        ^~~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/__config:897:17: note: expanded from macro '_VSTD'
#  define _VSTD std
                ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/vector:919:21: note: in instantiation of function template specialization 'std::allocator_traits<std::allocator<RMGTrackOutputScheme::RMGTrackEntry>>::construct<RMGTrackOutputScheme::RMGTrackEntry, int, int, int, int &, int, double, double, double, double, double, double, double, double, void, void>' requested here
    __alloc_traits::construct(this->__alloc(), std::__to_address(__tx.__pos_),
                    ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/vector:1678:9: note: in instantiation of function template specialization 'std::vector<RMGTrackOutputScheme::RMGTrackEntry>::__construct_one_at_end<int, int, int, int &, int, double, double, double, double, double, double, double, double>' requested here
        __construct_one_at_end(std::forward<_Args>(__args)...);
        ^
/Users/zichenwang/remage/src/RMGTrackOutputScheme.cc:100:17: note: in instantiation of function template specialization 'std::vector<RMGTrackOutputScheme::RMGTrackEntry>::emplace_back<int, int, int, int &, int, double, double, double, double, double, double, double, double>' requested here
  fTrackEntries.emplace_back(
                ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/include/c++/v1/__memory/construct_at.h:39:38: note: candidate template ignored: substitution failure [with _Tp = RMGTrackOutputScheme::RMGTrackEntry, _Args = <int, int, int, int &, int, double, double, double, double, double, double, double, double>]: no matching constructor for initialization of 'RMGTrackOutputScheme::RMGTrackEntry'
_LIBCPP_HIDE_FROM_ABI constexpr _Tp* construct_at(_Tp* __location, _Args&&... __args) {
                                     ^
1 error generated.
make[2]: *** [src/CMakeFiles/remage.dir/RMGTrackOutputScheme.cc.o] Error 1
make[1]: *** [src/CMakeFiles/remage.dir/all] Error 2
make: *** [all] Error 2`

311Francis:
I have also did some ctest. Most of the test that fails concerns with the fact that I don't have hdf5 downloaded on my laptop which does not allow remage to use lh5 output. Except those, here are some things that fails. 
        114 - particlefilter/basicfilter (Failed)               extra
	115 - particlefilter/killvolume (Failed)                extra
	116 - particlefilter/keepvolume (Failed)                extra
	117 - particlefilter/killprocess (Failed)               extra
	118 - particlefilter/keepprocess (Failed)               extra
	120 - python/pytest-wrapper (Failed)                    extra mt
	121 - trackoutput/gen-output (Failed)                   extra
	122 - trackoutput/test-output (Not Run)                 extra
	125 - observables-ge/run (Failed)                       extra
	126 - observables-ge/tracks-bulk-no-limit (Not Run)     extra
	127 - observables-ge/tracks-bulk-10um-limit (Not Run)   extra
	128 - observables-ge/observables (Not Run)              extra

311Francis:
comparison
difference

How to turn IB on:

`/RMG/Processes/EnableInnerBremsstrahlung true
`
add the macro command before initialize the run. 

I have reduced my geometry a little bit. 
1. I have removed the coax and the LAr sphere since they take too much computation power and the root file also records too much data that my laptop does not have room to save.
2. In the end, I have a BEGe at the center in an empty world, and I randomly generate a Ar39 decay within the world outside the detector. 

Above is the detector response. 
1. There is IB gamma and generated. The difference of turning IB on and off centers around 100-150 keV which is consistent with my previous simulation on Geant4.
2. The general detector response is different between remage simulation and my own simulation. One important difference is that I did not have LAr around my detector in remage simulation but I have LAr in my own simulation at Geant4.
3. There are also other aspects that remage has with a simple Geant4 simulation, which might also cause detector response differences.

Some errors and warnings after turning IB on:

```
[Detail --> RMG_IB_GenericIon: Inner Bremsstrahlung wrapper process initialized
[Error ---> fmt library exception caught: argument not found
[Detail --> RMG_IB_triton: Inner Bremsstrahlung wrapper process initialized
[Error ---> fmt library exception caught: argument not found
[Detail --> Register cuts for G4Region ‘SensitiveRegion’
```

`[Warning -> encountered long global time (> 285.616 yr). Global time precision might be worse than 1 us.`

311Francis:
The simulation above has a total of 10^7 decay for each simulation

tdixon97:
Thanks Francis. Some of the issues seem to be related to your system (I think). Definitely push your updates and we can review it directly 
The validation plot is a nice start, i think you should run the simulation in LAr not vacuum however, that will actually change a lot since in vacuum there is no (outer)-Bremstralhung. It would also be better to save lh5 files and plot this with python, especially for the validation report.
I am surprised the CPU time and memory is so long, be sure you are only recording steps in Ge and only when the energy is > 0 keV (you can find remage macro commands to do it). I also suggest making some truth level plots using the track output scheme. 
Let me know if you need help with how to do that!

311Francis:
I just pushed the newest version with both the error message and warning of decay time fixed.

311Francis:
I have a question on the detector signal, there are way more low energy deposition in my remage simulation comparing with my pure Geant4 simulation. I am wondering if there are some energy threshold or signal processing that remage has but my Geant4 does not have?

311Francis:
Screenshot 2025-04-10 at 10 33 45 AM
Above is my simulation result on my previous Geant4 simulation. I am wondering if remage have some production cut? Since the remage detector signal is significantly more repressed on the low energy end comparing with my pure Geant4 simulation which just looks for the total energy deposition of events.

tdixon97:
2GB is not too large, I recommend not saving hits in the Scintillator output this will lighten it alot

tdixon97:
I really doubt the vertices can be some GB, how many primaries did you generate? Maybe you can send me the macro

311Francis:
Screenshot 2025-10-06 at 1 49 41 PM
I did really increase the number of primary events, which is still 10^7, but I did reduce my sample sphere's radius from 15cm to 6 cm. Above is the new graph which is very similar to my previous Geant4 Simulation. I will move on to create the track output file.

tests/processes/plot_IB_effect.py
```diff
@@ -0,0 +1,100 @@
...
+    fig, ax = plt.subplots(figsize=(10, 3))
+    h_IB_rebinned.plot(yerr=False, label="With IB (20 keV bins)")
+    h_noIB_rebinned.plot(yerr=False, label="Without IB (20 keV bins)")
+
+    plt.ylabel("counts / 20 keV")
+    plt.legend()
+    plt.savefig("Detector_signal_comparison.output.png")
+
+    # Plot the rebinned difference
+    fig, ax = plt.subplots(figsize=(10, 3))
```
ManuelHu:
```suggestion
    plt.subplots(figsize=(10, 3))
```

tests/processes/plot_IB_effect.py
```diff
@@ -0,0 +1,100 @@
...
+
+plt.rcParams["lines.linewidth"] = 1
+plt.rcParams["font.size"] = 12
+plt.rcParams["figure.figsize"] = (10, 3)
+
+style = {"yerr": False, "flow": None, "fill": True, "lw": 0.6, "alpha": 0.5}
+
+
+def plot_vertex_energy():
+    fig, ax = plt.subplots()
```
ManuelHu:
```suggestion
    plt.subplots()
```

ManuelHu:
@311Francis I allowed myself to squash together your commits and to add some fixes:
* from clang-tidy
* minor code-style tweaks
* make the validation tests actually run and work in the validation suite

one minor thing remainging: please add license headers to the two C++ files you created. you can copy it from another file, update name and year and add your ORCID (if you have one).

tdixon97:
Thanks @311Francis for the great work! Unfortunately the tests fail since there are overlaps in the gometry

311Francis:
I just added the new licence headers to the files I created. Is there anything else that I need to work on for this update?

gipert:
Thank you Francis, we should review this PR after the meeting next week.

311Francis:
> > Also for the physicslist.md, should I add documentation to the "optional component" section?
> 
> yes, for example.
> 
> > So is there any way that I can improve it? Since I was trying to mimic the python script from other tests
> 
> the script is not the problem and the tests run fine after I wired up them into the test infrastructure in my recent changes. The problem is more the physics :) it just requires a lot of statistics to see a significant change due to inner bremsstrahlung that is not on the order of statistical fluctuations. I already tried to improve this in one of my commits, this was not enough without more sophisticated filtering/biasing to extract the effect in a more visible way...

I think in my macro I have put 10^7 or 10^8 and there is around 3.5% of differences. And in my original Geant4 simulation, I put even more statistics to see that 3-4% differences. I think in general Inner Brem is a correction with this amount of magnitude.

311Francis:
> Could you check if there's a command like `/process/em/setBiasingFactor` that you can already use to bias the x-section?

I don't think so. I also checked the macro command list on the remage page, and I also did not see anything related to it.

tdixon97:
Ah I think the idea of Luigi was just to use biasing for this process with a dedicated macro command. Most other processes support this with inbuilt geant4 commands. The idea is to make sure we have the same support for our custom processes.

311Francis:
> Ah I think the idea of Luigi was just to use biasing for this process with a dedicated macro command. Most other processes support this with inbuilt geant4 commands. The idea is to make sure we have the same support for our custom processes.

That is actually where I was ocnfused about. I tried to call the piasing command through 
```
***** COMMAND NOT FOUND </process/em/setBiasingFactor 100 true> *****
***** Batch is interrupted!! *****
```
and later I also tried specify the biasin process, region, and energy like 
```
***** COMMAND NOT FOUND </process/em/setBiasingFactor compt BEGe 100 100 MeV true> *****
***** Batch is interrupted!! *****
```
and all of them have failed.

ManuelHu:
I think we can conclude this more-or-less off-topic discussion on the builtin  `/process/em/setBiasingFactor` command here. It will not affect this process, as the biasing would need to be implemented somewhere (it is not just magic that just happens)

btw.: both your invocations of the command did not follow the upstream documentation on how to use the command. For me the command works fine after `/run/initialize` (before, the physics list is not present).

---

but I have an idea on how to implement useful biasing here, I will push that later

docs/validation/processes-ib.md
```diff
@@ -0,0 +1,90 @@
...
+
+Internal Bremsstrahlung is a radiative correction to beta decay where a photon
+is emitted along with the beta particle and neutrino. Unlike external
+bremsstrahlung (which occurs when the beta particle interacts with matter after
+emission), internal bremsstrahlung is emitted directly from the nucleus during
+the decay process itself.
+
+The IB photon energy spectrum is continuous, extending from zero up to the
+Q-value of the decay. For Ar-39 (Q = 565 keV), the IB photons are typically soft
+(low energy), with most photons below 0.5 MeV.
```
gipert: This should go to the manual, in the part about the physics.

include/RMGPhysics.hh
```diff
@@ -144,6 +144,7 @@ class RMGPhysics : public G4VModularPhysicsList {
     bool fUseOpticalCustomWLS = true;
     bool fUseNeutronThermalScattering = false;
     bool fUseGrabmayrGammaCascades = false;
+    bool fUseInnerBremsstrahlung = false;
```
gipert: i think this should be true.

tdixon97:
I would like to see the performance difference, @311Francis can you extract that?

311Francis:
> I would like to see the performance difference, @311Francis can you extract that?

Screenshot 2025-12-03 at 7 42 30 PM

Above the difference with and without Inner brem after the update on the biasing factor. The biasing is factor of Inner Brem is 10 in this situation.

include/RMGInnerBremsstrahlungProcess.hh
```diff
@@ -0,0 +1,164 @@
...
+  public:
+
+    /**
+     * @brief Constructs a new RMG Inner Bremsstrahlung wrapper process.
+     *
+     * @param aNamePrefix Prefix for naming the process (default "RMG_IB").
+     * @param aType Process type (default @c fDecay).
+     */
+    explicit RMGInnerBremsstrahlungProcess(
+        const G4String& aNamePrefix = "RMG_IB",
```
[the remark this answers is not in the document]
311Francis: So should I leave it as it is? Since I was wondering if there is any other issues that I need to fix.

ManuelHu:
From my side this can be merged like this. Please have a look again @gipert. it is **disabled by default** right now. Some issues still have to solved, but these all require more testing with real workloads:

* the effect of the prefix of the process name, or whether we can omit this prefix
* the runtime performance impact in various scenarious
* how is the interplay with our step grouping. Now we can have the case where multiple low-energy gammas are emitted directly from the decay process

tdixon97:
Presumably a process like Bremsstrahlung with IR divergence should respect the chosen production cuts?

PR #479, opened 2025-09-11
add Calorimeter detector and output scheme

### LINKED ISSUE #489: Calorimeter detector and output scheme
We need a very simple detector that just logs the total energy deposited in a sensitive volume.

gipert:
I'm a bit unsatisfied by the fact that we still have a jagged energy structure in the output, even if I would like to store only the total energy deposited. I'm not sure we want to implement an exception for this scheme though.


gipert:
This should be now good, I just need to add some tests

python/remage/utils.py
```diff
@@ -1,7 +1,17 @@
...
+from collections import defaultdict
+
 
 def _to_list(thing):
     if not isinstance(thing, tuple | list):
         return [thing]
     return thing
+
+
+def _icp_to_dict(thing):
```
ManuelHu, line 12: shouldn't we move this to ipc directly?

gipert: what do you propose?

ManuelHu: I would add it on IpcResult (similar to get)

src/RMGCalorimeterDetector.cc
```diff
@@ -0,0 +1,103 @@
...
+    auto _hit = new RMGDetectorHit();
+    _hit->energy_deposition = 0;
+    fHitsCollection->insert(_hit);
+  }
+  auto* hit = static_cast<RMGDetectorHit*>(fHitsCollection->GetHit(0));
+
+  hit->physical_volume = pv;
+  hit->detector_uid = det_uid;
+  hit->energy_deposition += step->GetTotalEnergyDeposit();
+  hit->global_time = prestep->GetGlobalTime();
```
EricMEsch, line 96: I dont understand the logic here. So there is only ever 1 hit in the Calorimeter hit collection?

And this hit sums the energy, but keeps only the physical volume, uid and time of the last hit?

So what is even the use of tracking the last physical volume etc.?

gipert: yeah good point, we might just keep all values and store the lowest time at the end.

tdixon97:
This feature was requested again by some colleagues working on gamma assay, is it easy to finish ?

gipert:
I would say no, it's a bit tricky. To me it's not clear exactly what kind of information is this output scheme storing and if it's really useful (at least for our application which was to just log the summed energy deposited in some parts)

EricMEsch:
I think the implementation in general is pretty easy, the main issue was how to do the python post-processing, right?

gipert:
uhm I don't get this failure... @ManuelHu?

ManuelHu:
> uhm I don't get this failure... @ManuelHu?

me neither. Apparently it writes different data from unrelated output schemes, but only on Geant4 11.4.... I will investigate a bit

gipert:
I think this is how we want it, could you please review @legend-exp/remage-software?

gipert:
did you identify the origin of the regression? it's not related to this PR right?

ManuelHu:
It is somehow related: it does not happen when no calorimeter detectors are present in the geometry. (i.e. whenm I comment out the calo* volumes in the test  gdml)

edit: and it already happens with the HDF5 output, so it is not in the python code.

ManuelHu:
hopefully it now works

gipert:
thanks Manuel, now time to review this!

src/RMGCalorimeterOutputScheme.cc
```diff
@@ -0,0 +1,175 @@
...
+  if (hit_coll->entries() <= 0) {
+    RMGLog::OutDev(RMGLog::debug, "Hit collection is empty");
+    return;
+  } else {
+    RMGLog::OutDev(RMGLog::debug, "Hit collection contains ", hit_coll->entries(), " hits");
+  }
+
+  auto rmg_man = RMGOutputManager::Instance();
+  if (rmg_man->IsPersistencyEnabled()) {
+    RMGLog::OutDev(RMGLog::debug, "Filling persistent data vectors");
```
ManuelHu:
```suggestion
    RMGLog::OutDev(RMGLog::debug_event, "Filling persistent data vectors");
```

EricMEsch:
Looks good to me. Just an idea: it might be interesting to have two times? The time of first energy deposit and time of last energy deposit? Just because from an implementation perspective it would be fairly simple to add now. But i am not sure how useful that would be.

I guess you could at least see if the energy deposit was a point-like event or multiple deposits from different sources

PR #494, opened 2025-09-26
confinement: fix sampling from volumes with daughters

### LINKED ISSUE #493: sampling bug with daughter volumes
When sampling from multiple physical volumes, where these volumes are nested. Consider `fiber_cl2` -> `fiber_cl1` -> `fiber_core` and sampling from `fiber_*`. This selects all three volumes.

However, the weight is calculated from the _solid_'s CubicVolume, not the LV's. This means that all three volumes have approx. the same volume, leading to the same weight

ManuelHu:
fixes #493 

I think a similar correction for the surface area is not necessary, as we always just sample on the outside of the mother volume? @tdixon97 

### before
left volume with daughter cube as a larger point-density close to the surface

with-daughters output-before

### after
now the two a perceptually similar

with-daughters output-after

### DIFF
--- a/docs/manual/confinement.md
+++ b/docs/manual/confinement.md
@@ -159,6 +159,8 @@ All sampling modes described above are available, with few notes/limitations:
   uniform vertex surface density.
 - Sampling on the surface of volumes that contain daughters will result in
   vertices being distributed on the outer surface of the mother volume _only_.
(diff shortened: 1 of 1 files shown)


ManuelHu:
I added a paragraph on that

PR #519, opened 2025-10-28
confinement: add option to weigh by mass instead of volume

ManuelHu:


tdixon97:
I think for the background model we also want for some sources to give a different weighting, I am thinking of 2vbb where you want this proportional to mass*enrichment.
Easy to add?

PR #522, opened 2025-10-29
add option for multi-processing

### LINKED ISSUE #520: implement pseudo-multithreading with multiple processes
This is important to "fix" the multithreading issue (#287 ) for the simflow.

We should be able to handfle eventid-offsetting in the C++ code (pass a process index from python, and add `index*expected _events` to the evtid in the output scheme.

important: this would change the total amount of events simulated by a factor of the thread count when compared to multithreaded simulations.

ManuelHu:
this passes the old and new tests. I do right now not test the continuity of the evids between the files... (that would actually be desriable I guess...?)

fixes #520

### DIFF
--- a/docs/manual/output.md
+++ b/docs/manual/output.md
@@ -93,6 +93,9 @@ remage -o OUTPUT.lh5 -t 8
 
 will result in output files `OUTPUT_t0.lh5,..., OUTPUT_t7.lh5`.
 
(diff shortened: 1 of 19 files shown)


ManuelHu:
~~somehow the processes do not run in parallel on macOS... :-(~~ ah no, this is more complicated

PR #531, opened 2025-11-17
remove upper version bounds on LEGEND packages

gipert:
This will make it easier to install remage and other LEGEND packages in the same conda environment.

### DIFF
--- a/tests/observables/config/hit_config.yaml
+++ b/tests/observables/config/hit_config.yaml
@@ -54,7 +54,7 @@ processing_groups:
 
   - name: vtx
 
-    lh5_group: "/"
+    lh5_group: null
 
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -166,3 +166,36 @@ messages_control.disable = [
...
+[tool.pixi.dependencies]
+# build
+cmake = "*"
+make = "*"
+
+# run
+geant4 = ">=11.2.2"
+clhep = "*"
+xerces-c = "*"
+hdf5 = "*"
```
gipert: @ManuelHu this seems to install a non-threadsafe hdf5 version... do you understand why?

ManuelHu:
should be fine now

PR #572, opened 2025-12-26
G4 11.4 compatibility and test stable container version

gipert:


gipert:
It's `DOCKER_TAG` that gets set to `this`...

gipert:
This seems undocumented, I don't understand why it would be set to `this`. Anyways I guess we have to update the build hook to make it work for PR autobuilds.

ManuelHu:
I think the PR autobuilds were never really working (it always way building the latest code in main, not froim the PRs repo/branch). Also the PR builds should not push the :dev tag, I guess.

PR #585, opened 2026-01-13
Add combine_tracks validation

EricMEsch:
This is a first draft. I still have to write the actual validation markdown file. 

This validation might be overkill to just check a simple functionality but i still think it is useful. The validation 
- Checks primary gammas in a 2m lAr box with 0.5, 2 and 5 MeV 
- Checks that the energy deposited per event is within floating point precision for the same seed with `/RMG/Output/Scintillator/Cluster/CombineLowEnergyElectronTracks` turned on or off. (Right now the only way energy escaped is by leaving the 2m lAr box, there might be a better geometry for this)
- Plots the energy deposition vs location with extra information on the file-size and and number of unique tracks in the output (see below)

combine_tracks_edep_locs_5000keV

I will add a follow up PR to improve the `combine_low_energy_tracks` function for larger maps from ~4000 s to ~1s in my test with a map size of 400000 entries (but i still need to test different populations of low energy to high energy tracks etc).

For this i should probably also somehow add a time benchmark to this validation...

### DIFF
--- a/docs/validation/performance.md
+++ b/docs/validation/performance.md
@@ -2,7 +2,7 @@
 
 This section tests the performance of remage for critical functions and options.
 
-## Combine low energy tracks options
+## Combine low energy tracks option
 
 This part evaluates the performance of the
 `/RMG/Output/.*/Cluster/CombineLowEnergyElectronTracks` option. This option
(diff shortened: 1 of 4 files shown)


tdixon97:
why not you just make the box bigger?

PR #599, opened 2026-01-22
Add neutron validation runs

EricMEsch:
I still need to write the actual validation page.

But this would produce plots like this for a bunch of isotopes in a setup containing liquid argon inside a steel cylinder inside a watertank. There is also a moderator with Gadolinium inside of the Argon.
image

image

The simulation should have a dual purpose:

1.  The neutron starts with 5 keV, meaning there is Geant4 neutron transport involved. The plots show the number of captures on each isotope. While we can not compare this to "literature values" we can compare between remage versions. If the number of neutron captures on a certain isotope visibly changes, then there were changes regarding the neutron transport or capture cross sections.
2. The gamma spectrum after the neutron capture is shown. Also the Q-literature values according to the IAEA Nuclear Data section is shown. These gamma spectrums can then be compared to those of other publications. ( I have not added them directly, as first of all i do not know the rules of stealing gamma spectrums of other people and secondly that is way more work)

Both simulations use the same seed. But due to the different process the random engine gets out of sync almost immediately.

Are there any comments or proposals of what i might change? 
Some of the isotopes will have a very low statistic. But there is not really more feasible in the CI. Their plots will still be generated, as the total Q-value is still meaningful. Theoretically i could also simulate each capture individually in a world only made of the volume, but i think overall that takes more time for less information.


EricMEsch:
Is the surface containment test still flaky? And is it expected that it fails for both versions simultaneously?

ManuelHu:
> Is the surface containment test still flaky? And is it expected that it fails for both versions simultaneously?

yes, and yes... (today is your lucky day)

EricMEsch:
@ManuelHu do you know what is wrong now? The mac test failed to the usual confinement check. But the latest passed and the stable failed to some strange reason

ManuelHu:
It just could not download some necessary files because of a network error, not the fault of this PR. edit: the same with the rtd build failure

EricMEsch:
Okay, if this runs now the only thing left to do is make the simulation run in mt mode and then i would consider this PR done

EricMEsch:
I cleaned the commit history into 2 clean commits. One for the actual tests and the other one was adding the StoreVelocities option to the GermaniumOutputScheme.


Should be good to merge now?

ManuelHu (approved):
looks good for me

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
+# Neutron transport and capture spectrum
+
+Here we investigate the neutron capture physics. For this, there are four
+different setups.
+
+- Simple world filled with `"G4_STAINLESS-STEEL"`. The only active process for
+  neutrons is `nCapture` and neutrons are sampled at 10 keV. To compare the
+  de-excitation gammacascades against other models.
```
gipert:
```suggestion
  de-excitation gamma cascades against other models.
```

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
+# Neutron transport and capture spectrum
+
+Here we investigate the neutron capture physics. For this, there are four
+different setups.
```
gipert: maybe you could add links to the relevant remage docs

https://remage.readthedocs.io/en/stable/manual/physicslist.html#hadronic-physics-options

https://remage.readthedocs.io/en/stable/api/classRMGNeutronCaptureProcess.html#_CPPv424RMGNeutronCaptureProcess

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+
+- Simple world filled with `"G4_STAINLESS-STEEL"`. The only active process for
```
gipert: can we move these descriptions to the relative sections? i think it would make things clearer

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+  neutrons is `nCapture` and neutrons are sampled at 10 keV. To compare the
+  de-excitation gammacascades against other models.
+- Simple world, filled with `"G4_Gd"`. The only active process for neutrons is
+  `nCapture` and neutrons are sampled at 0.1 eV. To compare the capture
+  distribution and check that the _remage_ custom `RMGnCapture` process produces
+  the same result as the native Geant4.
+- Simple world, filled with `"G4_Gd"`. The only active process for neutrons is
+  `nCapture` and neutrons with a log-uniform distribution from 0.01 eV to 1 MeV.
+  To re-create the relative cross-section of nCapture on gadolinium and compare
```
gipert:
```suggestion
  To re-create the relative cross-section of `nCapture` on gadolinium and compare
```

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+  stable Argon isotopes. Every process is active. Neutrons are sampled with a
+  log-uniform distribution from 0.01 eV to 10 MeV. To re-create the total
+  cross-section for neutrons in argon.
+- Complex world, containing Argon, Gd-loaded moderator, Steel and Water.
+  Neutrons are sampled at 5 keV but every process is active. To validate in a
+  more realistic scenario.
+
+## Simple steel world
+
+Here, the gamma spectrum after the nCapture is shown. The goal is to validate or
```
gipert:
```suggestion
Here, the gamma spectrum after the `nCapture` is shown. The goal is to validate or
```

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+compare this against other models, which are missing for now. As previously
+mentioned: Simple world filled with `"G4_STAINLESS-STEEL"`. The only active
+process for neutrons is `nCapture` and neutrons are sampled at 10 keV.
+
+```{figure} ./_img/neutrons/neutron_spectrum_steel_shielding.output.png
+:width: 800px
+```
+
+Additionally a pie-chart showing the distribution of the nCaptures can be shown.
+This can be used to validate the nCapture cross-sections of steel at 10 keV
```
gipert:
```suggestion
This can be used to validate the `nCapture` cross-sections of steel at 10 keV
```
check formatting everywhere

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+  more realistic scenario.
+
+## Simple steel world
+
+Here, the gamma spectrum after the nCapture is shown. The goal is to validate or
+compare this against other models, which are missing for now. As previously
+mentioned: Simple world filled with `"G4_STAINLESS-STEEL"`. The only active
+process for neutrons is `nCapture` and neutrons are sampled at 10 keV.
+
+```{figure} ./_img/neutrons/neutron_spectrum_steel_shielding.output.png
```
gipert, line 49: use widths that are fraction of the page everywhere

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+:width: 800px
+```
+
+Additionally a pie-chart showing the distribution of the nCaptures can be shown.
+This can be used to validate the nCapture cross-sections of steel at 10 keV
+
+```{figure} ./_img/neutrons/neutron_pie_chart_steel.output.png
```
gipert, line 68: this figure is rendered quite big, can you set it to 50% maybe?

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+:width: 800px
+```
+
+## Geant4 nCapture vs RMGnCapture
```
gipert:
```suggestion
## Geant4 `nCapture` vs `RMGnCapture`
```

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+:subcaptions: above
+
+:::{image} ./_img/neutrons/neutron_pie_chart_normal.output.png
+:width: 90%
+:alt: Distribution using the Geant4 nCapture.
+:::
+
+:::{image} ./_img/neutrons/neutron_pie_chart_rmgncapture.output.png
+:width: 90%
+:alt: Distribution using the custom RMGnCapture.
```
gipert:
```suggestion
:alt: Distribution using the custom `RMGnCapture`.
```

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+  against literature.
+- Three simple worlds, filled with liquid Argon that consists only of one of the
+  stable Argon isotopes. Every process is active. Neutrons are sampled with a
+  log-uniform distribution from 0.01 eV to 10 MeV. To re-create the total
+  cross-section for neutrons in argon.
+- Complex world, containing Argon, Gd-loaded moderator, Steel and Water.
+  Neutrons are sampled at 5 keV but every process is active. To validate in a
+  more realistic scenario.
+
+## Simple steel world
```
gipert: Maybe better sth like "captures on steel"?

docs/validation/neutrons.md
gipert: i don't understand the offsets of the histograms in this plot

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+:width: 90%
+:alt: Distribution using the Geant4 nCapture.
+:::
+
+:::{image} ./_img/neutrons/neutron_pie_chart_rmgncapture.output.png
+:width: 90%
+:alt: Distribution using the custom RMGnCapture.
+:::
+
+If both plots do not show the same thing (within std) there is an issue.
```
gipert: what std? what is std?

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+
+We can also take a look at the de-excitation gamma spectrum:
+
+```{figure} ./_img/neutrons/neutron_spectrum_gd_shielding.output.png
+:width: 800px
+```
+
+## Relative cross-section of gadolinium
+
+First we take a look at the latest G4NDL data that is stored for Geant4, to see
```
gipert, line 110: is this taken from the actual geant4 installation the validation suite is run with?

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+```{figure} ./_img/neutrons/neutron_gd_cross_section.output.png
+:width: 800px
+```
+
+## Argon transport
+
+First, the total cross-section for neutrons with the different argon isotopes is
+compared. For this, Geant4 typically uses G4NDL data. In the following plots,
+the latest G4NDL data is directly (dynamically) extracted out of the Geant4 data
+folder within the remage container. The solid lines shows the static JEFF-3.3
```
gipert:
```suggestion
the latest G4NDL data is directly (dynamically) extracted out of the data
folder part of the Geant4 installation. The solid lines shows the static JEFF-3.3
```

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+missing for certain isotopes. This is indicated by the label stating "missing:
+...". The letters stand for : "C" means the "Capture" data is missing, "E" means
+the "elastic scattering" data is missing and "I" means the "inelastic
+scattering" data is missing.
+
+```{figure} ./_img/neutrons/neutron_jeff_vs_g4ndl.output.png
+:width: 800px
+```
+
+To not only verify the datafiles, but also that they get applied correctly and
```
gipert:
```suggestion
To not only verify the data files, but also that they get applied correctly and
```

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+the simulation works, the total cross-section for neutrons in liquid Argon is
+reconstructed from an idealized simulation. In this idealized simulation an orb
+of 2 km radius is used. There are three worlds and three simulations, one for
+each argon isotope. The density (and temperature) of the argon is that of liquid
+argon. To calculate the total cross-sections, first the steplength between two
```
gipert:
```suggestion
argon. To calculate the total cross-sections, first the step length between two
```

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+steps is derived, using the position of each step. The mean of this value is
+taken as mean free path and used to derive the total cross-section. The energy
+is given by the velocity of the particle at pre-step. Due to the 2 km radius of
+the orb, geometric boundaries should not be relevant. Additionally, a distance
+to surface check only counts those events, that end up at least 10 m away from
+any boundary. Still, the results look like there might be a bias from the
+simulation. This could be caused by non-physical simulation processes limiting
+the steplength, but has not been investigated so far.
```
gipert: add a TODO admonition box to say that this should be investigated with future studies

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+argon there is also a neutron moderator, which consists of hydrogen and is
+loaded with gadolinium. The goal of the geometry is to enable not only neutron
+capture on different isotopes, but also have the neutron transport be a crucial
+component regarding capture distribution. The main elements on which can be
+captured in this geometry are therefore H, Ar, Cr, Fe, Ni, Gd.
+
+First, the pie-chart showing where the captures took place is shown:
+
+```{figure} ./_img/neutrons/neutron_pie_chart_general.output.png
+:width: 800px
```
gipert: also very big figure, reduce to 50% width

docs/validation/neutrons.md
```diff
@@ -0,0 +1,271 @@
...
+`G4PhotonEvaporation` model to be used by specifying the macro command
+`/process/had/particle_hp/use_photo_evaporation true`. The `G4PhotonEvaporation`
+model should always correctly respect the Q-value of the capture, but is very
+likely to not correctly represent the gamma line intensity. The standard
+`G4Shielding` option might fall back to the `G4PhotonEvaporation` depending on
+the available data and version. Both plots might show a different number of
+captures per isotope due to the random engine getting out of sync. Still numbers
+should be within standard deviation of another.
+
+Only these two models are shown due to run-time constraints. Due to the
```
gipert, line 274: is this really the maximum stats we can have on the validation suite runner in a reasonable amount of time? @ManuelHu

gipert (dismissed):
thanks a lot Eric, i left some suggestions on how to improve the text

one request: i think this validation suite should always clearly tell the user what they are supposed to check in each plot / how the plots are supposed to be used. you do that sometimes but sometimes i think it can be clarified. could you check again?

docs/validation/neutrons.md
ManuelHu: unfortunately we do not have a good way to link back, so this is a bit fragile. We cannot use the typical references as in the rest of the documentation... It might also be a bit bad that this would always link to the stable docs, and not to the version the validation suite is built for?

docs/validation/neutrons.md
EricMEsch: The offsets are inspired by the plot grabmayr sent to me and i think it looks good to just have all isotopes in one singular plot. As the total height of the histogram does not matter, it is only about the intensities. Between every isotope there is an offset of 1e5. That number was just arbitrarily chosen by eye, regarding what looks best.

Bigger number obviously means more empty space between lines, which in turn means less height for each isotope (if you keep the figure dimensions).

gipert: it's ok but it's not explained anywhere atm?

docs/validation/neutrons.md
EricMEsch: yes, i clarified it in the text

docs/validation/neutrons.md
EricMEsch: This test went from ~250s to ~110s on my local pc after the multi-threading change. The runtime did not significantly change here on the runner after that change (i checked and it actually took longer on mac for example).

This tells me that this validation part is not significantly contributing to the runtime. So i can increase the statistics. But it gets problematic quick i guess.

It should be quicker after the option from Moritz PR to kill all secondaries in the stepping. But after that i do not think there is any way to speed it up more

docs/validation/neutrons.md
```diff
@@ -0,0 +1,291 @@
...
+[IAEA Nuclear Data Section](https://www-nds.iaea.org/) is shown in each plot for
+the respective isotope. This can directly be compared with the orange line,
+depicting the summed energy of all secondary particles resulting from the
+neutron capture process in Geant4. In blue, the gamma line intensity is shown.
+
+For every isotope there are always two plots shown. The plot on the left just
+specifies the basic `G4Shielding` physicslist to the simulation without any
+extra modifications. The plot on the right instead additionally forces the
+`G4PhotonEvaporation` model to be used by specifying the macro command
+`/process/had/particle_hp/use_photo_evaporation true`. The `G4PhotonEvaporation`
```
MoritzNeuberger, line 264: You might also want to test for `/process/had/particle_hp/skip_missing_isotopes true` and `/process/had/particle_hp/do_not_adjust_final_state true` since they are highly recommended by the [G4ApplicationBook](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/TrackingAndPhysics/physicsProcess.html?highlight=adjust#high-precision-neutron-interactions-neutronhp)). When the former is not set (default), G4 will use a isotope similar to the missing one insted which we definetly want to avoid. When the latter is not set (default), G4 will emit additional gammas to preserve energy conservation. For the latter, there is an argument to keep it set to false, though.

docs/validation/neutrons.md
```diff
@@ -0,0 +1,291 @@
...
+
+Here we investigate the neutron physics. _remage_ allows for the specification
+of hadronic physics with the macro command
+<project:../rmg-commands.md#rmgprocesseshadronicphysics> as explained in
+<project:../manual/physicslist.md#hadronicphysicsoptions>. This validation will
+focus on the `Shielding` option for hadronic physics. Additionally, _remage_
+allows for a modification of the neutron capture process according to
+<project:../manual/physicslist.md#customgrabmayrgammacascades>, which will also
+be validated. For this, there are five different main parts.
+
```
MoritzNeuberger, line 39: I think it would be useful for general audience readers to provide a very brief overview of the neutron physics implementation in G4, i.e., what the idea behind Neutron_HP/Particle_HP is and what G4NDL and evaluated nuclear data libraries are. Also, maybe a quick reminder about the relevant processes for neutrons in matter (elastic/inelastic scattering, radiative capture) and how their cross sections look (e.g., you reference the resonance-area below without explanation). This information is already scattered throughout the document, but I think it would be better to explain it in the introduction section.

Also, maybe a bit of motivation on why we are interested in validating this with a reference to L1000, and why it is interesting to understand the materials you tested.

docs/validation/neutrons.md
EricMEsch: ah ok

docs/validation/neutrons.md
EricMEsch: What change do you have in mind? I would like to avoid adding even more simulations. These tests already consist out of 9 simulations and every added sim takes statistics from the others (to stay reasonable in time).
For the first one i think we can make an argument to set it to true just by remage default? 
Does it affect the gamma cascades or also the capture cross-section?

MoritzNeuberger: It should only affect the cross sections:
> /process/had/particle_hp/skip_missing_isotopes true
>
>    This UI commands sets to zero the cross section of the isotopes which are not present in the neutron library. If Geant4 doesn’t find an isotope, then it looks for the natural composition data of that element. Only if the element is not found then the cross section is set to zero. On the contrary, if this variable is not defined, Geant4 looks then for the neutron data of another isotope close in Z and A, which will have completely different nuclear properties and lead to incorrect results (highly recommended).

I think it would be useful to check whether G4 actually sets the cross section to 0 and doesn't nick it from a completely unrelated isotope. Maybe that might be also what we are seeing with Ar36 and Ar38?

For the second command, it is a question of whether we want to conserve individual events' energy (false) or ensure that the statistical results are consistent with G4NDL (true). The false state is probably more what we want.

Regarding the statistics: I thought that with the kill-all-secondary implementation, the statistics would no longer be an issue. You may also be able to get more out of the multiprocessing or multithreading of individual simulations (we have 32 cores on the server after all).

docs/validation/neutrons.md
ManuelHu: another comment on this section: references to the normal documentation(e.g. with `<project:../` will not work in the validation suite.

EricMEsch: Okay, i added a small abstract to the introduction part. But i did not go into the relevant processes for neutrons in matter, as i kinda feel like the validation suite is not the place to explain basic physics... I might be wrong idk

MoritzNeuberger: It would be nice to have that, or at least pointers to where people can read more about it, i.e., links to the relevant pages in the G4 documentation.

docs/validation/neutrons.md
```diff
@@ -0,0 +1,320 @@
...
+as explained in the
+[Hadronic Physics Options](https://remage.readthedocs.io/en/stable/manual/physicslist.html#hadronic-physics)
+section. This validation will focus on the `Shielding` option for hadronic
+physics. Additionally, _remage_ allows for a modification of the neutron capture
+process according to
+[this section](https://remage.readthedocs.io/en/stable/manual/physicslist.html#custom-grabmayr-gamma-cascades),
+which will also be validated.
+
+The materials tested in these validations are chosen based on their relevance
+for particle physics experiments, such as LEGEND-1000.
```
MoritzNeuberger, line 38: On that note, if we argue that it is relevant to muon-induced isotope production, we should also test the cross sections for O and C in PMMA and (maybe significantly more relevant) for Ge in HPGe. 

In the future, we should also overlay recent experimental data from https://www-nds.iaea.org/exfor/ to compare with both simulation/G4NDL to check consistency.

MoritzNeuberger: I should clarify that these do not necessarily have to be implemented in this PR, just as a note for the future.

docs/validation/neutrons.md
```diff
@@ -0,0 +1,320 @@
...
+The Q-Values of the neutron capture process is well known for all captures in
+this setup. The expected Q-value according to the
+[IAEA Nuclear Data Section](https://www-nds.iaea.org/) is shown in each plot for
+the respective isotope. This can directly be compared with the orange line,
+depicting the summed energy of all secondary particles resulting from the
+neutron capture process in Geant4. In blue, the gamma line intensity is shown.
+
+For every isotope there are always two plots shown. The plot on the left just
+specifies the basic `G4Shielding` physicslist to the simulation without any
+extra modifications. The plot on the right instead additionally forces the
```
MoritzNeuberger: Rather than saying it is G4Shielding, specify that it is the Neutron_HP model within Shielding. Since it is also used by all other relevant hadronic physics lists, this makes clear that it is not a feature of just the Shielding physics lists, but of this specific model and therefore default physics list independent.

MoritzNeuberger: Maybe also add some references here on the specific gamma cascade models.

docs/validation/neutrons.md
```diff
@@ -0,0 +1,320 @@
...
+use the cross-section of similar isotopes instead.
+
+## Complex geometry
+
+Last, we try to show some parameters simulating a more complex world, similar to
+many particle physics experiments. Neutrons are sampled at 5 keV but every
+process is active. Neutrons are simulated in a steel container filled with
+liquid argon. The steel container itself is located within a water volume and
+inside of the argon there is also a neutron moderator, which consists of
+hydrogen and is loaded with gadolinium. The goal of the geometry is to enable
```
MoritzNeuberger, line 224: I think it would be nice to have a small render of the geometry to better understand the setup and the relative expected contribution of the individual volumes' isotopes.

docs/validation/neutrons.md
```diff
@@ -0,0 +1,320 @@
...
+The plot shows the total weighted cross-section for natural argon according to
+JEFF-3.3, G4NDL and using this reconstruction method.
+
+```{figure} ./_img/neutrons/neutron_argon_summary.output.png
+:width: 70%
+```
+
+Because some isotopes might be missing in G4NDL, it can be possible that the
+simulation does not reproduce the G4NDL prediction. It might therefore be useful
+to investigate the cross-sections for the individual isotopes again:
```
MoritzNeuberger, line 191: Consider adding a sentence here that in G4NDL4.7.1 the cross-sections for Ar36 and Ar38 were removed and that their disagreement with JEFF is expected. Maybe also add a reference to the entry in the change logs: https://geant4.web.cern.ch/download/release-notes/notes-v11.2.2.txt

src/RMGGermaniumOutputScheme.cc
```diff
@@ -148,6 +155,10 @@ void RMGGermaniumOutputScheme::AssignOutputNames(G4AnalysisManager* ana_man) {
       CreateNtupleFOrDColumn(ana_man, id, "zloc_post_in_m", fStoreSinglePrecisionPosition);
       CreateNtupleFOrDColumn(ana_man, id, "dist_to_surf_post_in_m", fStoreSinglePrecisionPosition);
     }
+    if (fStoreVelocity) {
```
MoritzNeuberger, line 151: This is more of a note for the future and not a note to this PR specifically: I suggest we rethink our approach to output schemes a bit. I am worried that specialized output schemes will become overly complex. I would be in favor of creating a general output scheme that allows users to specify which parameters to return. In turn, specialized output schemes like germanium, scintillators, and optical should then really only have what is necessary for production.

EDIT: I should clarify, I am fine with the way it is solved here. IMO, we should consider that in the future.

tests/neutrons/data/156Gd-100Entries.txt
MoritzNeuberger, line 1: Out of curiosity, would it also be possible to sideload these files from the paper? It's not necessary, it might just be a fun way for people to find out that there are additional generated files in the paper's appendix/zenodo. Alternatively, just give a heads-up in the text that these files exist.

docs/validation/neutrons.md
```diff
@@ -0,0 +1,320 @@
...
+:width: 50%
+```
+
+## Geant4 `nCapture` vs `RMGnCapture`
+
+In this simple world filled with `"G4_Gd"`, the only active process again is
+either `nCapture` or `RMGnCapture`. The neutrons are sampled at 0.1 eV. The
+pie-charts show the capture distribution on the gadolinium isotopes. They are
+expected to show the same thing in both cases, as the `RMGnCapture` process
+should only impact the gamma cascade and not the capture cross-section.
```
MoritzNeuberger: Consider giving a quick reminder on why we have a wrapper for the nCapture process to motivate this test.

src/RMGGermaniumOutputScheme.cc
ManuelHu: yes, but that also would require quite some different architecture, i.e. with creating instances of anoutputscheme from macros, and not only changing the parameters of a single global instance. Quite a lot of work, I think.

docs/validation/neutrons.md
```diff
@@ -0,0 +1,320 @@
...
+typically uses G4NDL data. In the following plots, the latest G4NDL data is
+directly (dynamically) extracted out of the data folder part of the Geant4
+installation. The solid lines shows the static JEFF-3.3 data (on which at least
+G4NDL4.7 is based on), while the corresponding dots (squares) show the latest
+G4NDL data. It can be possible, that G4NDL data is missing for certain isotopes.
+This is indicated by the label stating "missing: ...". The letters stand for :
+"C" means the "Capture" data is missing, "E" means the "elastic scattering" data
+is missing and "I" means the "inelastic scattering" data is missing.
+
+```{figure} ./_img/neutrons/neutron_jeff_vs_g4ndl.output.png
```
MoritzNeuberger, line 154: Consider scaling the cross-sections of the isotopes by their isotope fraction, such that it is clearer that the total corresponds to the sum of the individual contributions. 

Alternatively, if you want to highlight the cross sections of the individual isotopes separately, consider splitting this into four subplots, each comparing the JEFF and G4NDL data.

docs/validation/neutrons.md
```diff
@@ -0,0 +1,320 @@
...
+pre-step. Due to the 2 km radius of the orb, geometric boundaries should not be
+relevant. Additionally, a distance to surface check only counts those events,
+that end up at least 10 m away from any boundary. Still, the results look like
+there might be a bias from the simulation. This could be caused by non-physical
+simulation processes limiting the step length.
+
+:::{todo}
+
+Investigate the effect of non-physical simulation processes on this
+cross-section reconstruction.
```
MoritzNeuberger: Idea on how to cross-check (not necessary for this PR):

Set up a thin target neutron beam experiment and count how many neutrons are unchanged in energy and direction on the other side. This way, you could extract the total cross-section by scanning through energy.

docs/validation/neutrons.md
EricMEsch: I did not want to do this, as this validation suite should just run in the background for every remage and Geant4 version. So i did not want to add static information to the text, which will be outdated soon. Instead, i tried to make it visible in the plots, which are dynamicly produced. I mean thats the whole purpose of this validation suite.

If we just wanted to staticly check G4NDL4.7.1 that is a whole different thing

src/RMGGermaniumOutputScheme.cc
EricMEsch: Well i already started with a PR for this general output scheme once. And people did not want it

tests/neutrons/data/156Gd-100Entries.txt
EricMEsch: Interesting question. Although the files in the papers appendix are relatively big i think? So i am not sure we want to cause that download every single time the validation suite is run (which means every single time someone makes a commit to a PR).

The files are mentioned in the [manual](https://remage.readthedocs.io/en/stable/manual/physicslist.html#custom-grabmayr-gamma-cascades).

ManuelHu: if we set this up in a good way, there are also caching possibilities in github actions (also on our runner, I guess): https://github.com/actions/cache?tab=readme-ov-file

docs/validation/neutrons.md
EricMEsch: I think C and O is negligible? At least from experience i have barely seen anything happen on C and O (which is biased as that might be due to Geant4 being wrong, lol) (Also PMMA and with that C and O is in the more complex test). But it might be reasonable to also add a test for this. But i am somewhat always very deterred from flooding the validation suite with tests.

So maybe this is also a different question on whether we want the validation suite to focus only on core aspects or if we want it to be as extensive as possible.

Regarding Ge i totally agree. But i think for that case we might want to put more effort into it than just a basic comparison? (As i think you have done a lot of work, scaling it to other data etc.).

Anyways i am going to add a todo to the file, as i do not plan on adding more tests to this PR

docs/validation/neutrons.md
EricMEsch: I added the idea to the todo

docs/validation/neutrons.md
EricMEsch: I do like this idea. I did adjust it and it looks better i think

EricMEsch:
I applied changes for every comment i marked as "resolved" and let those open for which i did not apply any changes.

@ManuelHu right now i just added a picture for one of the setups as Moritz wanted (not sure if it works the way i did it). But i think there probably also is a way to just make a picture using `legend-pygeom-vis` from the commandline?

src/RMGGermaniumOutputScheme.cc
ManuelHu: I think Moritz' idea was a bit different: so not an extra GeneralOutputScheme, but just one main Output Scheme that is instanced and configured multiple times, i.e. once for germanium, once for scintillator (now they are becoming very close again), once for "...".

ManuelHu:
> But i think there probably also is a way to just make a picture using legend-pygeom-vis from the commandline?

sure there is. We never did this in the remage repo, but you could get some inspiration from here: https://github.com/legend-exp/legend-pygeom-l200/blob/main/docs/generate_docs_images.py

you can also adapt this for CLI use (essentially write a scene file with the necessary options) or do it in the python script. This works without X11 server.

gipert:
🥳 💯 

image

docs/validation/neutrons.md
EricMEsch: I think that probably is what we are seeing with Ar36 and Ar38, although i think they removed the warnings?
In [this post](https://geant4-forum.web.cern.ch/t/why-does-the-neutronhp-model-in-geant4-v11-2-2-give-warnings-about-missing-files-for-argon-isotopes-when-this-issue-did-not-occur-in-geant4-v11-2-1-and-v11-0-3-and-how-can-it-be-resolved/12786) it shows that.

Well, we will see the impact of the kill-all-secondary command. Multithreading the individual simulation might help currently (as the neutron simulations run 500 sec out of the total 699.01 sec. ). But this also comes with overhead again. And i think if there are going to be just more simulations, then just having each simulation run single threaded but in parallel (the way it is done now) is better.

docs/validation/neutrons.md
MoritzNeuberger: I guess it depends on who the validation suites should be written for and how they should be written. I understand that you would prefer it written so experts can quickly identify errors without having to read much boilerplate. If this were a purely technical document, I would agree with you.

However, I think it's better to consider a more "general" G4 programmer audience that doesn't know how hadronic/neutron physics is implemented in G4. Yes, some experts will review the tests to ensure everything is fine, but non-experts will do so as well, and they may raise unnecessary alarms. Additionally, a validation suite is the perfect place to educate people on these subjects, as understanding them is necessary to know what to expect. Therefore, we should not expect people to know how Neutron_HP works. They would also do not know where to find the cross-sections or why some of them might be missing. Spelling out these things would help them understand the expected disagreement. 

Finally, if not adjusted in the future, this kind of transient information about 4.7.1 would indeed be problematic, but since we will still have experts who will check these tests, it should be simple to remove the mention if the cross-sections are added again. It could be included in a warning section that would be difficult to overlook.

But again, this depends on one's opinion on how to write this documentation. If you are fine with it being a technical page only, then this is fine.

tdixon97: I agree @MoritzNeuberger but I think we should consider @EricMEsch point about the information becoming outdated. We definetly want the validation suite to have nice explanations but we should keep an eye on the future maintenance of this report. In principle you could check the version of G4NDL and add extra information if version == 4.7.1

EricMEsch: The issue with checking the version is that i can only do dynamic stuff in the python code, and the only current way to propagate this information is through the PNG that is created, which contains the plot.

AFAIK there is no other way currently, to relay dynamic information to the validation suite? As the text is static. Maybe you could add stuff to the `CMakeLists` but that would make it way more complicated again.

EricMEsch: I added a note about the isotopes missing now

ManuelHu:
Is there still anything crucial missing here?

EricMEsch:
Should be good from my side (except maybe cleaning the commit history)

gipert:
@EricMEsch could you tidy up the history and resolve all conversations? Then we can check one last time that everyone is happy and merge.

EricMEsch:
Should be good now.

gipert:
can you also rebase on main?

PR #620, opened 2026-02-12
validation infrastructure and updates

ManuelHu:


tests/CMakeLists.txt
```diff
@@ -1,12 +1,31 @@
...
 
+# ---------------------------------------------------------------------
+# calculate a statistics boost factor depending on the machine.
+if(RMG_VALIDATION_STATS_FACTOR)
+  set(STATS_FACTOR ${RMG_VALIDATION_STATS_FACTOR})
+else()
+  include(ProcessorCount)
+  ProcessorCount(_num_cores)
+
+  math(EXPR STATS_FACTOR "${_num_cores} / 8")
```
gipert: will this factor be low enough?

ManuelHu: It will be 4 on kali, and 1 on any machine with <= 8 cores (e.g. my laptop, current GitHub runners). As cmake is doing integer math, only at 16 cores you will get a factor of 2. I think this should be fine?

PR #627, opened 2026-02-18
feat: optimization of the electron stacking

MoritzNeuberger:
This pull request introduces optimizations to the recently added stacking scheme for electron tracks in the bulk of volumes. It optimizes safety calculations using caching and bounding spheres, and allows filtering to only consider Germanium detectors for certain checks. 

**Updates to the User Commands**
- The `/RMG/Output/VolumeStacker/` command group was adjusted. Instead of activating the filter for only a single volume, the user can add multiple volumes via the name `AddVolumeName` (register multiple volumes). In addition, the safety calculation can be adjusted to only consider the distances to HPGe detectors via `DistanceCheckGermaniumOnly` (limit checks to Germanium detectors). 

**Performance Improvements: Safety Calculation**
- Added an efficient `is_within_surface_safety` function that quickly determines if any surface is within a safety distance, with early exits and bounding sphere optimizations.
- Introduced a caching mechanism (`volume_cache` and `VolumeCache` struct) for volume geometry data.
- Introduced a simple precondition for the safety calculation via bounding spheres for daughter volumes, to speed up repeated distance checks.
- Introduced option to only check safety distances to HPGe detectors (see update to user commands).
- In case the safety is set to 0, the safety calculation is skipped, and all electrons in the volume are directly put on the waiting stack.

These changes were presented in an internal presentation on 26.01.2026 [[link](https://indico.legend-exp.org/event/2409/)]. Since then, some more tests have been performed. The previous tests were run with the multithreading option, which turned out to distort the mean time per event, and systematic differences were introduced due to differences depending on when the simulations were run. Now the simulations are all run single-threaded and at the same time. The plot below shows the mean time per event for a Tl208 simulation in L1000 for different E_low thresholds and different configurations of safety compared to the optical photon stacking implementation and w/o optical physics.

image

A slight increase in the mean time per event for larger safeties is visible, but they are all still significantly smaller than with the only optical photon stacking. 

EDIT: I realize that I probably should rebase the branch to the main repo. Will do that in a bit.


include/RMGEventAction.hh
```diff
@@ -40,6 +41,10 @@ class RMGEventAction : public G4UserEventAction {
   private:
 
     RMGRunAction* fRunAction = nullptr;
+
+    // Timing statistics for event processing
+    std::chrono::high_resolution_clock::time_point fEventStartTime;
+    std::chrono::nanoseconds fEventTotalTime{0};
```
ManuelHu: why is this a class member. you are using this only once?

src/RMGOutputTools.cc
```diff
@@ -34,6 +34,26 @@
 
 #include "magic_enum/magic_enum.hpp"
 
+namespace {
+  // Configuration for bounding sphere optimization
+  bool g_filter_germanium_only = false;
+
+  // Cache structure for volume geometry data
+  struct VolumeCache {
```
ManuelHu: please rebase this on main, as there the VolumeCache is already present (in RMGNavigationTools, so a different location!). You should probably do that as soon as possible to avoid duplicate work

also, why do you have a struct declaration in a non-header file?

src/RMGStackingAction.cc
```diff
@@ -27,8 +27,22 @@
...
+
+  RMGLog::Out(
+      RMGLog::debug_event,
+      "Looping over output schemes for track ",
+      aTrack->GetTrackID(),
+      " of particle type ",
+      aTrack->GetDefinition()->GetParticleName()
+  );
   for (auto& el : fRunAction->GetAllOutputDataFields()) {
+    RMGLog::Out(RMGLog::debug_event, "Asking output scheme ", typeid(*el).name());
```
ManuelHu: do we really need this debug logging here? what value does it give the user?

MoritzNeuberger:
There was a presentation on 13.4.2026 for a status update on this PR [[link](https://indico.legend-exp.org/event/2493/)]. 

A point of interest discussed was that the differences in the 511 keV line are likely due to positron stacking. I removed the functionality, and the agreement is now again better than before (see figures at the end of the post).

**Plan for the PR:**
- [ ] refactor the implementation to clean up clutter accumulated over the long development time (including the above points)
- [ ] implement tests to cover functionality and expectation in optimization / physics
- [ ] write documentation detailing intended usage

A question I had was whether we should also do a **"light refactor" of the staging condition logic**. At the moment, it is necessary to set `/RMG/Output/Germanium/DiscardPhotonsIfNoGermaniumEdep true` so that electrons are also stacked. This is not straightforward nomenclature. Maybe use a consistent naming scheme, including `/RMG/Output/IsotopeFilter/DiscardPhotonsIfIsotopeNotProduced`? Maybe something like `RegisterStagingCondition`?




**Before**:
image

**After**:
image

gipert:
I agree with changing command name but `RegisterStagingCondition` sounds too generic

I also think it's a good idea to ask the AI to simplify!

MoritzNeuberger:
shit, didnt want to merge yet

EDIT: yes, I am aware that I pressed auto-merge ... I was convinced it was about rebasing to main ...

MoritzNeuberger:
Reverted the changes in main. I guess I have to make a new PR?

PR #629, opened 2026-02-26
output: store number of events in LH5 file

### LINKED ISSUE #615: Number of primaries
We should store the number of generated primaries either in the remage files or those from reboost. It is easy to get from the `vtx` table but people can easily make mistakes.

ManuelHu:
not fully implemented yet in the post-processing, but we should be able to discuss it on this draft already.

fixes #615

### DIFF
--- a/src/RMGConvertLH5.cc
+++ b/src/RMGConvertLH5.cc
@@ -307,9 +307,11 @@ bool RMGConvertLH5::ConvertToLH5Internal() {
   auto ntuples = GetChildren(ntuples_group);
   bool ntuple_success = true;
 
-  std::string links_group_name = "__by_uid__";
+  const std::string links_group_name = "__by_uid__";
+  const std::string n_ev_name = "number_of_events";
   std::vector<std::string> links;
   RMGIpc::SendIpcNonBlocking(RMGIpc::CreateMessage("lh5_links_group_name", links_group_name));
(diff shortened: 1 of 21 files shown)


src/RMGConvertLH5.cc
```diff
@@ -307,9 +307,11 @@ bool RMGConvertLH5::ConvertToLH5Internal() {
   auto ntuples = GetChildren(ntuples_group);
   bool ntuple_success = true;
 
-  std::string links_group_name = "__by_uid__";
+  const std::string links_group_name = "__by_uid__";
+  const std::string n_ev_name = "__number_of_events__";
```
gipert: I would not use the __, in the end we will use this a lot

gipert:
missing: a note about this in the manual section about output.

also i find `number_of_events` not so clear as a name. which events? maybe `number_of_primaries` is clrearer? or `nr_sim_events`?

cc @tdixon97

ManuelHu:
`number_of_primaries` would be misleading, as an event can contain more than one primary (i.e. from bxdecay0 or other external event generators)

gipert:
maybe then we should clearly state that these are simulated events? i think with the current name an user could think that these are detected events

ManuelHu:
yeah, I will add some docs in my pull request

PR #638, opened 2026-03-12
Add muon physics tests

MoritzNeuberger:
As highlighted in [link](https://github.com/legend-exp/remage/pull/599#issuecomment-3790797128), IMO, there is interest in adding a simple test on the muon physics. Specifically, I had these in mind:

- [x] Test 1: Regarding energy loss, have muons go through water and measure the deposited energy. The radiative loss should be a peak at lower energies. Stochastical loss should then be related to showers and should be at tail to higher energies. This should be tested with different em and hadr physics lists.
- [x] Test 2: Regarding muon-induced neutron yield, again have a simple setup as above and shoot muons at several energies and count the number of neutrons (ofc w/o double counting). Would have to be run to the end (i.e., until the capture or decay of neutrons), since secondary neutrons can be dominant, as in LAr. This should also be tested for several materials. This should be tested with different em and hadr physics lists, as well as different ENDL.
- [x] documentation/entry in the validation report

The first test is implemented since it is relatively straightforward. The second one needs more care since it might potentially require significantly more calculation time.

In addition, I added an option to the stepping action to kill any secundaries via `/RMG/Processes/Stepping/KillSecondaries true`. This makes implementing the first test significantly easier. If it is not desired, I can remove it again.


tests/cosmogenics/muon/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+  macros/*.mac gdml/*.gdml *.py)
+
+# copy them to the build area
+foreach(_file ${_file_list})
+  configure_file(${PROJECT_SOURCE_DIR}/${_file} ${PROJECT_BINARY_DIR}/${_file} COPYONLY)
+endforeach()
+
+# run python tests using pytest
+add_test(NAME cosmogenics/muon/pytest-all COMMAND ${PYTHONPATH} -m pytest -s -vvv .)
+set_tests_properties(cosmogenics/muon/pytest-all PROPERTIES LABELS "extra;val")
```
ManuelHu: you might consider setting also the label `val-only` if the tests take long to execute.

MoritzNeuberger:
Seems like the validation runs out of time after 25min, so I would have to reduce the runtime to meet that threshold. Is it possible/desirable to increase the timeout threshold?

ManuelHu:
you can increase it for each test: https://cmake.org/cmake/help/latest/prop_test/TIMEOUT.html#prop_test:TIMEOUT

EricMEsch:
I mean in general this whole run-time optimization is something very important to us, as it can directly be applied to our simulations for L1000. As in both cases our main concern is to not bias the statistics

docs/validation/muon_physics.md
```diff
@@ -1,4 +1,218 @@
...
+
+The energy loss of muons in matter consists of four main processes: ionization,
+bremsstrahlung, pair production, and photonuclear interactions. The first
+process is continuous and dominant at low energies, while the latter three are
+stochastical and dominant at high energies. The transition between the two
+regimes is around 100 GeV in water and liquid argon.
+
+In Geant4, ionization is implemented using three different models according to
+the energy of the muon
+[[Muon Ionisation]](https://geant4.web.cern.ch/documentation/pipelines/master/prm_html/PhysicsReferenceManual/electromagnetic/muon_incident/muion.html#mu-ionis),
```
ManuelHu: these are the wrong links. please use the ones from the websites, and not the pipeline-version, they might sometimes break in unpredictable ways...: https://geant4-userdoc.web.cern.ch/UsersGuides/PhysicsReferenceManual/...

(yes, I also do not understand why search engines seem to like the pipeline version so much more)

docs/validation/muon_physics.md
```diff
@@ -1,4 +1,218 @@
...
+a function of the muon energy. The simulation estimates were generated by
+shooting mono-energetic muons through a block of the material with a length each
+adjusted such that the muon would lose around 5% of its energy. In the plot, the
+simulated values are compared to the tabulated values from the PDG. The
+agreement between the points and the total expected should be good for the whole
+energy range.
+
+:::{image} ./\_img/cosmogenics/muon/energy_loss_water_energy_range.output.png
+:height: 400px :alt: Energy loss of muons in water compared to DOI:
+10.1006/adnd.2001.0861 :::
```
ManuelHu: you might need to use backtick fences here, not colon fences. Otherwise prettier will destroy the markup, as it happened here.

.pre-commit-config.yaml
```diff
@@ -83,11 +83,6 @@ repos:
         language: pygrep
         entry: PyBind|Numpy|Cmake|CCache|Github|PyTest
         exclude: .pre-commit-config.yaml
-      - id: disallow-end-fence
```
ManuelHu, line 86: please restore that check, end ::: fences not on their own line will not be picked up by the parser.

ManuelHu (commented):
for now, just some minor comments

tests/cosmogenics/muon/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+# run python tests using pytest
+add_test(NAME cosmogenics/muon/energy-loss COMMAND ${PYTHONPATH} -m pytest -s -vvv
+                                                   test_energy_loss.py)
+set_tests_properties(cosmogenics/muon/energy-loss PROPERTIES LABELS "extra;val;val-only" TIMEOUT
+                                                             1800)
+
+add_test(NAME cosmogenics/muon/particle-yield COMMAND ${PYTHONPATH} -m pytest -s -vvv
+                                                      test_particle_yield.py)
+set_tests_properties(cosmogenics/muon/particle-yield PROPERTIES LABELS "extra;val;val-only"
+                                                                TIMEOUT 360000)
```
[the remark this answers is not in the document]
ManuelHu: How long does it usually take at the moment, @MoritzNeuberger ?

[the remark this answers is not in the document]
MoritzNeuberger: The particle-yield test is ~20min, so 30min is ok. Maybe have it a bit more conservative at 1h?

docs/validation/muon_physics.md
```diff
@@ -0,0 +1,405 @@
...
+better understood, while muon spallation and secondary neutron multiplication
+dominate the modeling difficulty [^Wang]. Existing reference studies in FLUKA
+and Geant4 [^Wang] [^Kudryavtsev] [^Manukovsky] provide important guidance, but
+their numerical values depend on the chosen hadronic models and cross-section
+data and can be associated with large uncertainties.
+
+From the experimental side, data are scarce because neutron-yield measurements
+require large detectors with good containment and high detection efficiency.
+Water and liquid-scintillator experiments provide important constraints [^SK]
+[^SNO], but direct measurements for liquid argon at relevant muon energies are
```
EricMEsch: Why did you not mention MINIDEX? Just because it is located in Tübingen? :(
https://www.sciencedirect.com/science/article/pii/S0927650520301195

docs/validation/muon_physics.md
```diff
@@ -0,0 +1,405 @@
...
+
+In the simulation, mono-energetic muons are shot through a cylinder of material
+with a length of $2000 g/cm^2 \cdot \rho$ similar to [^Manukovsky]. The radius
+was not reported in [^Manukovsky], but was chosen to be 1.25 m to ensure good
+containment of the resulting showers. The neutron yield is then calculated as
+the number of neutrons produced per muon per g/cm$^2$ after the burn-in depth
+(see the shower dimensions section below).
+
+The following tests are useful to validate that the simulations are in the right
+ballpark and to check for large discrepancies between simulation and
```
EricMEsch: I am not complaining, but i find it hilarious, that this entire validation suite reads like a scientific paper and suddenly you drop the word "ballpark"

docs/validation/muon_physics.md
```diff
@@ -0,0 +1,405 @@
...
+parameterization [^Agafonova] based on experimental trends together with the
+material-specific Geant4-based study from [^Manukovsky] and the FLUKA
+simulations by [^Wang] and [^Kudryavtsev] for water. A key feature from
+[^Manukovsky] is that argon can deviate upwards from the generic trend of
+[^Agafonova] because of its low neutron separation energy and high $(n,2n)$
+cross-section.
+
+The following plot shows the muon-induced neutron yield in different materials
+at 100 GeV. The simulated values are compared to the general parameterization
+from [^Agafonova] and the material-specific results from [^Manukovsky]. The
```
EricMEsch: More or less the exact same sentence is already written on line 166-168

docs/validation/muon_physics.md
```diff
@@ -0,0 +1,405 @@
...
+ballpark and to check for large discrepancies between simulation and
+expectation.
+
+### Material dependence
+
+Following the power-law framework introduced above, this subsection focuses on
+the material dependence of the neutron yield. The comparison uses the
+parameterization [^Agafonova] based on experimental trends together with the
+material-specific Geant4-based study from [^Manukovsky] and the FLUKA
+simulations by [^Wang] and [^Kudryavtsev] for water. A key feature from
```
EricMEsch: Is the plot for the Water case missing? Or am i missing something?

Edit: It is in the subsection below?

docs/validation/muon_physics.md
```diff
@@ -0,0 +1,405 @@
...
+First, it is important to estimate the "burn-in" depth until the muon showers
+have properly developed, i.e., the average particle flux and thus the average
+energy deposition along the penetration depth stabilizes. This is relevant
+because in deep-underground simulations it is desirable to propagate muons
+through surrounding rock only as far as needed. The minimum simulation depth is
+therefore set by the burn-in depth after which the resulting particle flux in
+the cavern does not change significantly anymore. Since neutrons are the most
+relevant particles for LEGEND, we compare neutron production locations along the
+depth. The burn-in depth is material dependent and should be around 3-4 m in
+water, around 1-2 m in liquid argon, and around 1 m in rock. The burn-in depth
```
EricMEsch: I hate being that guy, but given that you cited everything else, from where are these values for the burn-in depth?

docs/validation/muon_physics.md
```diff
@@ -0,0 +1,405 @@
...
+
+```{image} ./_img/cosmogenics/muon/shower_dimensions_100.0_rock_Shielding_Livermore_30_8.045.output.png
+:height: 300px
+:alt: Muon shower profile in rock at 100 GeV.
+```
+
+```{image} ./_img/cosmogenics/muon/shower_dimensions_100.0_water_Shielding_Livermore_30_8.045.output.png
+:height: 300px
+:alt: Muon shower profile in water at 100 GeV.
+```
```
EricMEsch, line 496: I really like the plots, but can you add more about how you calculate the location of the stabilization? It looks like its just the first point at which the orange line crossed the red sigma band? As in this image it looks like due to a high statistical peak this happens too early: 

Image

EricMEsch: Also what exactly is the "neutron prod." value representing? The average number of neutrons? Or the sum over all showers? As i was not expecting it to produce on average like 100+ neutrons per bin (it looks like 12+ bins per meter. That would be a lot of neutrons per meter argon)

docs/validation/muon_physics.md
MoritzNeuberger: I didn't mention it here because the mean energy (7 GeV) and material (lead) are vastly different from those of interest for LEGEND. If there is interest in potential other experiments or setups for future versions of  remage, it might be worthwhile to extend the range to shallow depths and other materials.

docs/validation/muon_physics.md
MoritzNeuberger: Fair point. These are mainly guesses made from experience. I will add a sentence to mention this.

docs/validation/muon_physics.md
MoritzNeuberger: good catch!

docs/validation/muon_physics.md
MoritzNeuberger: There is only one plot for the material dependence, since I am considering only one slice in A at an energy of E=100 GeV. Water has an average A of ~14.3, rock has an average A of 22, and Ar has an average A of 40. I will consider adding labels to the plot.

docs/validation/muon_physics.md
MoritzNeuberger: After all, we're all still just children with expensive toys, aren't we? ;)

docs/validation/muon_physics.md
MoritzNeuberger: I tried to add an explanation on this in the docs. Please let me know if this is sufficient to understand the process.

EricMEsch: Yes, just an idea: How about normalizing the "neutron prod." axis to the number of muons? So it would be neutrons produced per muon. I think that would be an interesting metric?

MoritzNeuberger:
I guess some stuff broke with the new version of lgdo / lh5 being a separate package?

gipert:
yes, there are two separate problems. the deprecation warning will always make it fail, but there is an issue with the lgdo.lh5.exceptions module which can be fixed on the legend-pydataobj side

gipert:
moritz, can you add <2 to the pyproject for now?

MoritzNeuberger:
What?
```
 80/144 Test  #67: confinement-surface/relative-fractions-box ................***Failed    6.19 sec
Traceback (most recent call last):
  File "/__w/remage/remage/build/tests/confinement/./test_basic_surface.py", line 219, in <module>
    raise ValueError(msg)
ValueError: box has 1 in 500000 primaries not close to any side with absolute tolerance 1e-06 mm
```
Is this test known to be flaky?

ManuelHu:
>  Is this test known to be flaky?

yes (all of the `relative-fractions-*` tests)

tdixon97:
We should probably change the threshold, I have never understood this failures, it seems to me the test is ok, but sometimes the sampling is wrong

gipert:
see #462, but it was a bit weird since loosening up the threshold wasn't really fixing it

tdixon97:
Yeah, I think that there is some rare bug somewhere in the sampling, when I looked into it in the past it really did seem the returned vertex wasn't exactly on the surface. I updated #462 so lets discuss there

MoritzNeuberger:
This PR is ready for review (once again). The features are final IMO.

gipert:
can you resolve the conversations and cleanup the git history?

tests/cosmogenics/muon/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+
+# copy them to the build area
+foreach(_file ${_file_list})
+  configure_file(${PROJECT_SOURCE_DIR}/${_file} ${PROJECT_BINARY_DIR}/${_file} COPYONLY)
+endforeach()
+
+# run python tests using pytest
+add_test(NAME cosmogenics/muon/energy-loss COMMAND ${PYTHONPATH} -m pytest -s -vvv
+                                                   test_energy_loss.py)
+set_tests_properties(cosmogenics/muon/energy-loss PROPERTIES LABELS "extra;val;val-only" TIMEOUT
```
ManuelHu, line 14: some of these tests use multiple CPUs, right? When CMake does not now about that, it will try to schedule as many other tests in parallel, leading to overcommitment of resources and potential test timeouts.

you always need also to set the number of used cpu cores here, and (optionally) the mt label if the tests uses all cores.

tests/processes/selective_ekin_min.py
```diff
@@ -0,0 +1,163 @@
...
+    reg = create_geometry(volume_material)
+
+    gdml_path = tmp_path / f"geometry_{particle}_{cut}.gdml"
+    writer = pg.gdml.Writer()
+    writer.addDetector(reg)
+    writer.write(str(gdml_path))
+
+    macro_path = tmp_path / f"macro_{particle}_{cut}.mac"
+    macro_content = f"""
+/RMG/Manager/Logging/LogLevel detail
```
ManuelHu, line 40: writing temporary files is no longer necessary or recommended. all the subsituted values should work with the normal macro_substitutions feature.

gipert:
There are still some CI issues to solve...

ManuelHu:
yes, exactly:

* (again) the command file mismatch
* after merging `main` you have to adjust your new scripts to use the new lh5 import. I only silenced the DeprecationWarnings in external code.

also please clean-up the git history of the changes since yesterday...

ManuelHu (approved):

Issue #541, opened 2025-11-23
Depth of surface confinement


tdixon97:
Currently surface contamination can only be simulated **_exactly_** on the surface of the solid. However, there are many scenarios where the contamination should instead be ~nm - um deeper.
I think this will become important for modelling data after LAr veto since it changes the probability for beta/ alphas to escape from a surface.

It should be possible to implement by finding the normal vector to a solid (`G4VSolid::SurfaceNormal `) and then pushing the points a certain distance in that direction. We may want to implement a few discrete options (eg. exponential [link](https://arxiv.org/abs/2405.17937)) or something more Gaussian (expected from ion recoil implanation [link](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://indico.cern.ch/event/1199289/contributions/5445792/attachments/2704294/4695344/TAUP2023.pdf&ved=2ahUKEwi1isOqqIiRAxV1a0EAHRbmJLQQFnoECBIQAQ&usg=AOvVaw2CbpeMi2-rArbQpUmrvYIA).

PR #643, opened 2026-03-26
[WIP] Surface sampling depth

tdixon97:
Created by the AI so needs alot of work / improvements but essentially uses #541 . I think this is really important for modelling data after LAr (@rosannadeckert ?)


gipert:
This looks good to go for me, and I think we should mention this in the paper.

@ManuelHu can you take a look and merge?

tdixon97:
@gipert I think we need a validation plot, what I would do is generate this for eg. HPGe and plot the distance to surface and check it matches the supplied profile.

I think it fits well in the distance to surface or the confinement sections.

Can claude do it?

tdixon97:
share plots

gipert:
the validation pages are building

gipert:
a bit too fast manuel... we wanted to see the plots. here they are:

image
image

i'm actually not so happy about the last one, one can't really see

tdixon97:
for me they are ok

ManuelHu:
but the last plot also checks it numerically, that they are inside?

tdixon97:
Yeah @ManuelHu thats why i also think its fine

PR #645, opened 2026-04-01
use global transformation for distance to surface

### LINKED ISSUE #644: investigate distance_to_surface for volumes with complex positioning
the calculation of the transform of the global hit coordinates to the local object frame did not properly take into account the translations of the mother volume(s). This leads to wildly wrong values.

ManuelHu:
the updated distance validation tests should now also cover this case and fail if not implemented correctly.

fixes #644

### DIFF
--- a/src/RMGNavigationTools.cc
+++ b/src/RMGNavigationTools.cc
@@ -29,6 +29,12 @@
 
 #include "RMGLog.hh"
 
+/// \cond this triggers a sphinx error or creates weird namespaces @<long number>
+namespace RMGNavigationTools {
+  G4ThreadLocal std::unordered_map<const G4VPhysicalVolume*, VolumeCacheEntry> volume_cache;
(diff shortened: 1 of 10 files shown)


src/RMGNavigationTools.cc
```diff
@@ -268,4 +268,14 @@ std::vector<RMGNavigationTools::VolumeTreeEntry> RMGNavigationTools::FindGlobalP
   return trees;
 }
 
+RMGNavigationTools::VolumeTreeEntry RMGNavigationTools::FindGlobalPositionCached(
```
EricMEsch: Why is the function name `cached`? It just re-calculates the distance every time right?

ManuelHu: I think you commented on an outdated commit? in the latest version it is not called `Cached` any more...

Issue #628, opened 2026-02-19
Ideas about controlling optical physics processes via macros


MoritzNeuberger:
The motivation is that, for muon simulations, we can use optical maps in the LAr volumes, but there is no such thing in the water tank. Therefore, we need to track the optical photons in the water tank, but not in the LAr.

An idea brought up by @EricMEsch was to provide an option to individually turn off scintillation, so that only Cherenkov photons are produced. This would also produce Cherenkov photons in the LAr, though we won't care so much for them. Implementation wise, this would be straightforward by adding an additional boolean in `/RMG/Processes/` that allows the `scint_proc` initialization to be skipped. 

I was wondering if something like this should be part of a more general approach that gives users the option to turn off certain processes if needed, ie, for testing purposes.

ManuelHu:
I think processes per volume/region are also not really supported by Geant4?

ManuelHu:
> I was wondering if something like this should be part of a more general approach that gives users the option to turn off certain processes if needed, ie, for testing purposes.

1. Deactivating processes is already possible via (built-in) macro `/process/inactivate Scintillation all`
2. We could in principle also switch to the upstream optical physics constructor, that has more logical controls to enable/disable processes . (note: this is a bit of validation work, as the upstream constructor has a slightly different structure as our current one).

EricMEsch:
Regarding just the neutrons another option would be to again just use the track outputscheme to write out the nCapture info and then simulate them all in a second step.

But this still does not help you with the primary muon cerenkov response

MoritzNeuberger:
In any case, rather than controlling the optical physics in remage, this should be handled in the geometry generation process. Since the optical parameters are stored in the GDML, one could implement logic in those scripts to decide which parameters to include. I will close this issue, though the ideas about the water tank might be worth looking into in the future.

PR #654, opened 2026-04-14
Unified approach to staging of particles (previously: "feat: optimization of the electron stacking")

MoritzNeuberger:
Well, this is awkward... this is a second attempt at this PR after the first one (#627) was merged by accident. Sorry again for the trouble. Here we go again:

This PR restructures the already present staging in remage into a single, user-configurable `StagingScheme` class. This new scheme owns all defer-to-waiting logic, i.e., it decides whether a particle should be placed on the urgent or waiting stack. This replaces the previous split implementation where optical-photon deferral lived inside specific output schemes (Germanium and IsotopeFilter), removing code duplication in the process. IMO this is a much cleaner model with explicit separation between:

1. Deferring tracks to the waiting stack with options for users to tune the conditions for electrons (in `StagingScheme`)
2. Deciding whether waiting-stack tracks should be cleared at stage transition (in the `Germanium` or `IsotopeFilter` output schemes)

At the moment, the optical photon and electron staging logic must be activated individually. We might want to automatically activate the former when the latter is activated. 

Electron staging from #627 is preserved and integrated into the same central interface. Users can tune electron deferral with volume, safety, and energy controls to reduce performance/physics tradeoffs in boundary-heavy geometries. If no electron volume is specified, staging applies to all volumes.

Overview of new and redefined user commands

`/RMG/Output/ActivateOutputScheme Staging`

New staging controls:
```
/RMG/Staging/OpticalPhotons/DeferToWaitingStage
/RMG/Staging/Electrons/DeferToWaitingStage
/RMG/Staging/Electrons/VolumeSafety
/RMG/Staging/Electrons/MaxEnergyThresholdForStacking
/RMG/Staging/Electrons/AddVolumeName
/RMG/Staging/Electrons/DistanceCheckGermaniumOnly
```

Redefined condition-based waiting-stack controls:
```
/RMG/Output/Germanium/DiscardWaitingTracksUnlessGermaniumEdep
/RMG/Output/IsotopeFilter/DiscardWaitingTracksUnlessIsotopeProduced
```

What is still missing:
- [x] tests
- [x] auto-activate staging for optical photons when electron staging is activated 
- [x] gamma staging (of interest for muon simulations)
- [x] `SetTrackStatus(fSuspend)` option to defer tracks that, after tracking some time, satisfy the condition to be deferred
- [x] documentation
- [ ] validation page (expected performance improvement, visualizing expected discrepancies in expected signal)


gipert:
thanks moritz i agree that this seems cleaner!

MoritzNeuberger:
In addition to the staging applied to the track initialization (the ClassifyTrack call), I now also added the option to **suspend already running tracks to the waiting stack** when their energy drops below the energy threshold. This is especially of interest for muon simulations, where we want the stage transition condition to depend on the production of isotopes. We can apply the energy threshold below the production threshold for these particles (e.g., neutron separation energy), so they would not contribute to the production anyway. For this reason, **I also added the option to stage gammas**, which are in the same situation as electrons in this regard. IMO, this is a very nice feature, as it further reduces the number of optical photons stacked at the initial stage, reducing both memory usage and additional processing time. I will run some external tests to evaluate the performance improvement in muon simulations.

MoritzNeuberger:
Over the weekend, I ran a simple muon simulation with optical physics to test the impact of the staging options implemented above. The setup consists of a (h=5cm, r=2.5cm) LAr volume through which I shoot 100 GeV muons with optical physics. I then ran ~200 events for different configurations of the staging that consist of the following:
- particles stageing options:
  - no staging
  - only optical
  - electrons (w/ and w/o optical staging), ethr = 8 MeV
  - gammas + electrons (w/ and w/o optical staging), ethr = 8 MeV
- w/ or w/o suspension when particles kinetic energy falls below 8 MeV

The resulting event processing times and max memory usage per event are plotted below (once the raw distributions, once the means).

I admit I am a bit disappointed and confused by the results. The main difference is, as expected, whether we stage the optical photons or not. W/ optical photon staging, independeon of the option, we do get the expected speedup but also always get significant memory usage (cluster top left), while without optical photon staging, the electron or electron+gamma staging options are always compatible with the (yellow) non-staging option (cluster bottom right). W/ optical photon staging, there is a slight improvement with the electron/electron+gamma staging compared to the just optical photon staging (turquise), though the memory usage is still significant.   

My hope was that with the electron+gamma+optical photon staging, the ram usage would be sufficiently low to run. Overall though, it seems like muon simulations with directly optical photons enabled (compared to separate simulations like Eric mentioned before) still seems quite unfeasable with this approach unfortunatly, since for a L1000 setup, the necessary ram for a single event can easily cause the simulation to crash.

EDIT: I should clarify that the electron+gamma staging would still be interesting for non-optical sims of muons, since, from previous tests, I expect their suspension to lead to a significant speedup when searching for isotope production.

EDIT2: I forgot to mention in the initial post, that the second stage is always skipped in this setup.

Raw distributions:
image

Means and their uncertainty:
image

MoritzNeuberger:
I repeated the above tests using the L1000 geometry for a more realistic example for a performance comparison. In addition, I only activated the water to be optically active (by manually removing all optical properties in the GDML for now, but this could be implemented as an option in the geometry generation) and used the musun input. Such a setup + optical maps might be the most reasonable approach for us to minimize simulation time while keeping memory usage low.

In the plots below, you can see that the mean now has significantly lower memory usage, while the scatter plot shows that individual events can still use up to ~50Gb. To further test the feasibility of this setup, I will run a stress test on NERSC, increasing the number of parallel processing tasks and seeing how many fail.

Average + std of the memory usage/processing time:
image

Scatter plot of the individual events:
image

ManuelHu:
I have allowed myself to re-create the cmmit history of this branch. I tried to rebase, but this was not really feasible due to the high number of merges from main and changes in between. I also fixed the reported clang-tidy issues

gipert:
should we consider merging this and working on the validation page separately?

not sure if this PR is ready to go though, i haven't reviewed it

MoritzNeuberger:
I think the PR is ready for review IMO.

To finish up the discussion above about the optical muon simulations: I think the best option we have is to apply optical physics in the water with optical + estacking and use the maps in the LAr. However, there are some additional caveats to consider.

Below, you see the RAM usage and event count for the L1000 simulations above over time (100 evts). t=0 is defined as the time the run starts, meaning RAM usage before is during initialization (in my setup, ~1min), while the drop after all 100 evts were processed corresponds to after the run (writing + post-proc, in my setup, ~10sec). The RAM usage always increases whenever there is an event with a large waiting stack and does not reset even when manually clearing the waiting stack (not implemented atm, but would look like something like this in the event action `G4EventManager::GetEventManager()->GetStackManager()->clear()`).

image

Since we cannot avoid the fact that some of these events have ludicrous RAM requirements, the only way to avoid having several such processes running at the same time would be to reduce the number of events per process. To make this efficient, one would have to find a sweet spot between the number of events per process, the number of threads, and the total individual simulation time (initialization + run + after), while keeping the expected total RAM usage reasonably below the machine limit, such that it is unlikely that it would kill any of the processes.

It's not pretty, I admit that, though I do not think there is a more efficient method (at least in terms of processing time). 

That is to say, this is only true as long as we still want to properly reproduce the PMT hit locations in the water tank. If we do not care which PMT saw which photon, then we could probably also use an optical map in the water (see [[link](https://github.com/legend-exp/remage/issues/628)]).

MoritzNeuberger:
I admit the stacking of the gammas didn't bring the performance boost I hoped for (as far as I remember from months ago). Admittedly, I also do not see much else of a use case. Maybe we might want to get rid of it again to avoid unnecessary complexity?

gipert:
I think it's important we have a clear motivation for all the features, yes

gipert:
I'm fine with merging this PR but it's too much for me to review... could you spend some time with Claude to make sure that everything looks right? also do we have some minimal validation for the report?

ManuelHu:
The most important thing is to verify that the exiting _useful_ staging/stacking features (i.e. photons after veryfiyoing a germanium edep) still work. The distance stacking itself is already broken in `main` right now, so this PR can only make it better...

MoritzNeuberger:
This should (hopefully) be covered by the tests (at least as far as I remember ... man, this PR has been dragging on, sorry). Will ask Claude for some review.

MoritzNeuberger:
The result of a vibe-coding session is:
- found and fixed a logical bug when staging in only specific volumes selected
- removed the staging of gammas since there is no use case for it
- fixed tests that were broken by this

I also already squashed the history.

EricMEsch:
I have just managed to read through this ( I missed it before, sorry).

I have not yet looked at the code, but from the description this looks like a really nice implementation! I like your investigation. I have also a lot of experience with this, although i sadly never really documented my experience. My experience was mainly based on "Will this run with the 500GB on NERSC or will this crash it eventually", but i did it based of a >10^8 statistic, so i can guarantee you, i hit all of the edge cases that caused my stuff to crash. So what i can say from that experience:
(I always have the optical properties inside the cryo removed. So similar to you, only water is optical):

- As you already stated, just putting optical photons directly on the stack guarantees running out of RAM. You dont even need a rare event.
- Putting electrons/gammas on the stack without putting optical photons on the stack will **not** yield a big speed-up if you do not **disable cerenkov effect for muons**. The primary muon will create so much cerenkov that the secondary light has a negligible effect. So it would be very interesting to see all of your benchmark plots, but with Cerenkov effect disabled for muons (This has to be done where remage applies cerenkov to the particles i think? I am not sure there exists a G4 macro for this?).
- Like you already motivated to have this upper threshold: Just putting all electrons/gammas on the stack without the upper threshold will result in ~50% less Ge77 productions. 
- If you just put an upper threshold on electrons and globally put them on the stack, you might still run out of RAM, due to a ton of low-energy electrons being put on the stack (in rare very extensive events). So my solution was to **not** put electrons < 100 keV on the stack (as they can not do cerenkov anyways). I also had an extra condition for the watertank: I put all electrons > 250 keV on the stack if they are created directly in the water (even if above threshold). I did not see a significant impact on isotope production from that.
- I treated positrons the same as electrons. I did not check yet if your logic also treats positrons, but it should.

With all of that i think the simulation runs reasonable fast, close to the "no optical". The big downside is that you lose the prompt muon response. And actually we would kinda like to have this response. But i did not find a reliable way to get it fast and include that response

EricMEsch:
And yes, upon consuming the memory Geant4 will never give the memory back during runtime. But internally i think it does share (although i am unsure if it shares among threats, i assume not?). But that means: If you had your most intense event in the beginning, then your RAM would never grow afterwards (as you can see in your plot).

docs/manual/staging.md
```diff
@@ -0,0 +1,285 @@
...
+optical photons, such as far-away radiogenic, and cosmogenic simulations. In
+other cases, such as close-by sources, it may be preferable to stage only
+optical photons.
+
+## Commands
+
+### Optical-photon staging
+
+- `/RMG/Staging/OpticalPhotons/DeferToWaitingStage` enables deferral of optical
+  photons to the waiting stack during stage 0.
```
EricMEsch, line 134: If you are using the word "stage" here to describe what is happening, it might be reasonable to also describe what stage 0 is? And are there just two stages or more? Above you just say "Geant4 transitions to a new stage" but not much more

docs/manual/staging.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Electron staging
+
+- `/RMG/Staging/Electrons/DeferToWaitingStage` enables deferral of secondary
+  electrons **as well as optical photons**.
```
EricMEsch: I would prefer it to not force both options to be combined?
Like this you are just taking away options

docs/manual/staging.md
```diff
@@ -0,0 +1,285 @@
...
+- `/RMG/Staging/Electrons/VolumeSafety` sets a minimum distance to a Germanium
+  detector surface condition.
+- `/RMG/Staging/Electrons/MaxEnergyThresholdForStacking` limits deferred
+  electrons by kinetic energy.
```
EricMEsch: From my experience it might also be useful to have a min Threshold for Stacking... (electrons below the cerenkov threshold)

docs/manual/staging.md
```diff
@@ -0,0 +1,285 @@
...
+/RMG/Generator/Confine UnConfined
+/RMG/Generator/Select GPS
+/gps/particle gamma
+/gps/energy 2.6 MeV
+/gps/ang/type iso
+
+/run/beamOn 1000
+```
+
+### Gamma example 2: optical staging with IsotopeFilter skip condition
```
EricMEsch: Why is this named "Gamma example"? In the previous example i thought the name is because the generated primary was a gamma. But here the primary is a neutron. If the name is because "opticalphoton = gamma", i am not sure if it is smart here, to break the Geant4 logic to strictly separate "gamma != opticalphoton"?

src/RMGGermaniumOutputScheme.cc
```diff
@@ -373,28 +372,22 @@ void RMGGermaniumOutputScheme::StoreEvent(const G4Event* event) {
...
-  // we are only interested in stacking optical photons into stage 1 after stage 0 finished.
+  // Optical-photon defer-to-waiting behavior is handled by RMGStagingScheme.
   if (stage != 0) return std::nullopt;
-
-  // defer tracking of optical photons.
-  if (fDiscardPhotonsIfNoGermaniumEdep &&
-      aTrack->GetDefinition() == G4OpticalPhoton::OpticalPhotonDefinition())
-    return fWaiting;
   return std::nullopt;
 }
```
EricMEsch: No reason to have this function here then right? You can just remove the overwrite of the header and remove this function here? The possibility to add the overwrite in the future is still there, in case there is a new reason to overwrite the function of the virtual parent

src/RMGIsotopeFilterScheme.cc
```diff
@@ -64,28 +63,22 @@ bool RMGIsotopeFilterScheme::ShouldDiscardEvent(const G4Event* event) {
...
-  // we are only interested in stacking optical photons into stage 1 after stage 0 finished.
+  // Optical-photon defer-to-waiting behavior is handled by RMGStagingScheme.
   if (stage != 0) return std::nullopt;
-
-  // defer tracking of optical photons.
-  if (fDiscardPhotonsIfIsotopeNotProduced &&
-      aTrack->GetDefinition() == G4OpticalPhoton::OpticalPhotonDefinition())
-    return fWaiting;
   return std::nullopt;
 }
```
EricMEsch, line 72: Same here

src/RMGStagingScheme.cc
```diff
@@ -0,0 +1,194 @@
...
+
+std::optional<G4ClassificationOfNewTrack> RMGStagingScheme::StackingActionClassify(
+    const G4Track* aTrack,
+    int stage
+) {
+  if (stage != 0) return std::nullopt; // only apply staging logic in stage 0
+
+  if (aTrack->GetDefinition() == G4OpticalPhoton::OpticalPhotonDefinition()) {
+    return Classify_OpticalPhoton(aTrack);
+  } else if (aTrack->GetDefinition() == G4Electron::Definition()) {
```
EricMEsch, line 35: I actually never tested, but i just assumed that positrons will have a comparable impact to electrons in cosmogenic sims? But i might also be wrong

We might need to test if stacking positrons will improve perfomance?

src/RMGStagingScheme.cc
```diff
@@ -0,0 +1,194 @@
...
+
+  // Keep primaries unaffected, matching existing staging behavior.
+  if (track->GetParentID() == 0) return;
+
+  if (track->GetTrackStatus() != fAlive) return;
+
+  if (track->GetDefinition() != G4Electron::Definition()) return;
+
+  const bool suspend_on_drop = fSuspendElectronsOnEnergyDrop;
+  const double threshold = fElectronMaxEnergyThresholdForStacking;
```
EricMEsch, line 63: is this just for Shorter variable names?

src/RMGStagingScheme.cc
```diff
@@ -0,0 +1,194 @@
...
+
+  // Suspend exactly at threshold crossing to avoid repeatedly suspending tracks that always stay
+  // below threshold.
+  if (pre_energy > threshold && post_energy <= threshold) track->SetTrackStatus(fSuspend);
+}
+
+std::optional<G4ClassificationOfNewTrack> RMGStagingScheme::Classify_OpticalPhoton(
+    const G4Track* aTrack
+) const {
+  // Electron staging also implies optical-photon staging (see DeferToWaitingStage guidance).
```
EricMEsch, line 82: no likey

src/RMGStagingScheme.cc
```diff
@@ -0,0 +1,194 @@
...
+
+  // If volume names are configured, only apply electron staging inside those volumes.
+  // (If none are configured, electron staging applies to all volumes.)
+  if (!fElectronVolumeNames.empty()) {
+    const auto vol_name = volume->GetLogicalVolume()->GetName();
+    if (!fElectronVolumeNames.contains(vol_name)) return std::nullopt;
+  }
+
+  // stop if electron safety is not configured.
+  if (fElectronVolumeSafety < 0) return std::nullopt;
```
EricMEsch, line 117: Wait, this has to be a bug, right?

In your last muon example you do not explicitly set the VolumeSafety, right? And in that case you just never stack the electrons?

The default value of the variable is -1

docs/manual/staging.md
EricMEsch: e.g. You could still end up with other charged particles creating light in your setup. And this might still cause RAM issues (it did for me). So you just bite the nail and simulate any extra light anyways.

docs/manual/staging.md
```diff
@@ -0,0 +1,285 @@
...
+/RMG/Output/ActivateOutputScheme Staging
+/RMG/Output/ActivateOutputScheme IsotopeFilter
+
+/run/initialize
+
+/RMG/Staging/OpticalPhotons/DeferToWaitingStage true
+/RMG/Staging/Electrons/DeferToWaitingStage true
+/RMG/Staging/Electrons/MaxEnergyThresholdForStacking 8.0 MeV
+/RMG/Staging/Electrons/SuspendOnEnergyDrop true
+/RMG/Staging/Electrons/AddVolumeName world_vol
```
EricMEsch, line 240: Look, here you do not specify any safety

include/RMGStagingScheme.hh
```diff
@@ -0,0 +1,83 @@
...
+
+    std::optional<G4ClassificationOfNewTrack> Classify_OpticalPhoton(const G4Track* aTrack) const;
+    std::optional<G4ClassificationOfNewTrack> Classify_Electron(const G4Track* aTrack) const;
+
+    bool fDeferOpticalPhotonsToWaitingStage = false;
+
+    bool fDeferElectronsToWaitingStage = false;
+    double fElectronMaxEnergyThresholdForStacking = -1;
+    bool fSuspendElectronsOnEnergyDrop = false;
+    double fElectronVolumeSafety = -1;
```
EricMEsch: And here, default value = -1

docs/manual/staging.md
MoritzNeuberger: These are leftovers from when I had the gamma-staging implemented. I will delete them now.

MoritzNeuberger:
Implemented feedback and squashed commits. IMO ready for merge.

MoritzNeuberger:
Implemented a flag to also allow positron staging, reusing the conditions of electrons.

MoritzNeuberger:
Ready for merge

PR #655, opened 2026-04-24
add hades spectrum test

### LINKED ISSUE #639: Add validation with data from HPGe test stand
I think it would be not so hard to include a spectrum from HADES and simulate it. We have now all the remage-based infrastructure to do that easily.

@tdixon97 can you make a recommendation?

### LINKED ISSUE #656: Document allowed states
The macro commands doc has the "allowed" states listed which is very useful. But we realised that this is not explained and the G4 jargon is a bit obtuse (eg. idle)

I think we could add a little bit of info to the top of this page.

ManuelHu:
fixes #639, #656


tests/gespectrum/plot_gespectrum.py
```diff
@@ -0,0 +1,135 @@
...
+            ha_norm2,
+            bins,
+            fill=True,
+            color="#ff0000",
+            label="before QC scaled",
+            alpha=0.2,
+        )
+    ax0.stairs(h_norm2, bins, fill=True, color="#ff0000", label="Data scaled")
+
+    ax0.stairs(h_sim, bins, label="MC", color="#0b5394")
```
tdixon97, line 107: Why like this, not with `hist` ?

tdixon97:
Thanks @ManuelHu .
Few questions / comments.

* Minor but I would change a bit the style of this plot, maybe the red can have some alpha to help see the line on top clearer, also I guess the y-axis limit is not really correct
image

*  I wonder if plots 1 and 2 are both needed?
* Is the data/MC spectrum very low stats? I am suprised that there are no events > 2615 keV, I would usually like to see the reproduction of this region? I.e that the TCS peaks are correctly reproduced

tests/gespectrum/plot_gespectrum.py
ManuelHu: i could only use hist for plotting, but for nothing else? why should I use hist then?

ManuelHu:
> Is the data/MC spectrum very low stats? I am suprised that there are no events > 2615 keV, 

yes something has gone wrong there. not sure why, it should have ~10e4 counts in the 2614 keV peak. Then it also has counts above.

tests/gespectrum/plot_gespectrum.py
tdixon97: Sure just a choice, I just prefer the syntax to this mpl things like stairs

ManuelHu:
discussion points from today:
* geometry differences found between g4simple and legend-pygeom-hades, but not clear which is correct (e.g., the source geometry)
* low energy effect potentially TL model
* data only only available until ~3.1 MeV (end of ADC dynamic range)
* use an IC detector with "fake" geometry like in SSD?

tdixon97:
Thanks Manuel, I think this is very nice!

gipert:
@ManuelHu I also think we should improve plots and validation suite text in the next PR

PR #668, opened 2026-05-09
ci: add skip conditions for validation

ManuelHu:


ManuelHu:
I think we wanted to have something like nthis top reduce the unnecessary runs on the validation runner?

gipert (approved):

PR #670, opened 2026-05-10
tests: use tolerance in more surface confinement comparisons

ManuelHu:
change rules:

* $$c > const \quad\rightarrow\quad c > const - tol$$
* $$c < const \quad\rightarrow\quad c < const + tol$$

## more description

two observations from the original code:
* some comparisons did not use the tolarance at all (was that intended?)
* the tolarance actually decreased the amount of points on the edges of any surface. so increasing tol would also increase the statistical failure rate (as we observed in #462...) 

## original comment 

> https://github.com/tdixon97/remage/blob/37644db796633306737e0d877a26d45c422c6750/tests/confinement/test_basic_surface.py#L42-L46
>
> am I stupid or are we not taking into account the `tol` for some of these comparsions, i.e. `(np.sqrt(x**2 + y**2) < 60)`?? Similar for the "middle comparisons" in the boxes case.
>
> And `abs(z) < 30 - tol` is wrong, it should be `abs(z-30)<tol`, right? similar for others

_Originally posted by @ManuelHu in https://github.com/legend-exp/remage/issues/462#issuecomment-4380226585_

### DIFF
--- a/tests/confinement/CMakeLists.txt
+++ b/tests/confinement/CMakeLists.txt
@@ -176,53 +176,24 @@ set(_solids tubby box con uni sub trd)
 
 foreach(det ${_solids})
 
-  add_test(
-    NAME confinement-surface/gen-vis-macro-${det}
-    COMMAND /bin/sh -c
-            "sed 's/\${det}/${det}/g' macros/vis-surface.mac > macros/vis-surface-${det}.mac")
-
-  set_tests_properties(
-    confinement-surface/gen-vis-macro-${det}
-    PROPERTIES LABELS "extra;val" FIXTURES_SETUP confine-surface-vis-macro-${det}-fixture
-               FIXTURES_REQUIRED confine-surface-gdml-fixture)
-
   add_test(NAME confinement-surface/${det}-vis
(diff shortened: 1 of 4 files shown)


gipert:
@tdixon97 can you please check this?

ManuelHu:
I am currently running the tests on our compute server to check if they still fail, so far it looks good. My worries would actually be that they now got to lenient...

tdixon97:
I am not sure 1000 is enough iterations to see the issue, but lets hope

tdixon97:
Can we ask some AI to review this? I think it would be good at catching when these conditions are wrong

ManuelHu:
> Can we ask some AI to review this?

go ahead, I unfortunately lost my copilot access in the recent changes to education benefits wrt copilot.

also an update on the stress test: I ran 3000 iterations of each test (so ~18000 in total). Only the `con` test failed _once_: (1 in 500000 primaries not close to any side)

tdixon97:
@copilot can you check all the conditions in test_basic_surface.py

ManuelHu:
I let claude review it and it thinks its fine and fixes the bug.

tdixon97:
Great! thanks Manuel

gipert (approved):

Issue #539, opened 2025-11-21
Remove pyg4ometry dependency


gipert:
At the moment we depend on both Geant4 and pyg4ometry (through reboost), this makes our dependency tree quite complex (and we already see the consequence of this while maintaining the conda-forge feedstock).

Since remage does not need pyg4ometry, let's do the simplest thing: duplicate the few reboost pieces that we need in this package and drop that dependency. I would also not rely on `build_hit()` at all and just iterate over the output file with `LH5Iterator`. In the long term we can think about another package but this is not needed now.

ManuelHu:
I thought more about this. This does not really help for the conda build problem. Even though remage does not require the full reboost and pyg4ometry, the tests very much do so. We run all tests in the conda-forge CI because the reviewer there asked us to do so...

gipert:
But it's a bit stupid because the user does not need to run the remage tests... maybe we can just remove this part 👼🏻

ManuelHu:
More thoughts on this (I let claude have a take on this). Unfortunetaly at this stage it is not done with a "simple iterator", as we need the rather complex GLMIterator from reboost....

tdixon97:
I think that would be replaced, since it was fairly generic but actually doing a relatively simple task here
________________________________

gipert:
yes, i think it's not so much the code that we need. i would not simply copy over what we have in reboost, it can be simplified by a lot.

@ManuelHu I think if you copy over the functions from reboost as is and ask claude to remove the unnecessary features / refactor it should do a good job

ManuelHu:
see #674 (I did just do what you proposed). I still want to import some tests from reboost, to verify that the new functions work.

gipert:
This is not really possible since the geometry benchmarker relies on pyg4ometry.

PR #674, opened 2026-06-01
remove reboost post-processing

ManuelHu:
this is mostly written by claude, but strongly depends on thre upstream reboost code.

see #539 - but that is _not_ fixed by this PR


python/remage/post_proc.py
```diff
@@ -108,21 +107,15 @@ def post_proc(
...
-                time_window=time_window_in_us,
-            )
-
-            # use reboost to post-process outputs
-            reboost.build_hit(
-                config,
-                {},
+            # post-process outputs: reshape detector tables by time-grouping
+            # and forward auxiliary tables unchanged
+            build_hit(
```
gipert: could we rename this function to something else, more explicit about what it does?

gipert:
looks good to me

ManuelHu:
this is also not enough to remove the pyg4ometry dependency, as the geombench python wrapper also uses it directly.

gipert:
oh interesting. i guess it's hard to remove that dependency there...

ManuelHu:
yes, but maybe we can make it somehow optional?

gipert:
i'm not sure it's a good idea, we want it in the conda forge build?

ManuelHu:
I think we now should have a sufficient test base here?

gipert (approved):

gipert:
Manuel I think this is good to go, but before or after #479?

ManuelHu:
I'd say after (we have to do a complex rebase in both cases)

gipert:
luckily there's Claude for this...

ManuelHu:
now it should also contain the code for the calorimeter output.

